PR #12, opened 2019-11-21
fix list format for nestedtable

kylemccollom:


src/docs/reference/introduction/page-arguments.mdx
```diff
@@ -1,4 +1,4 @@
----
+t---
```
jeesunikim: t!!! @kylemccollom

PR #21, opened 2019-12-16
More empty files, add missing orders

vcarl:


src/api/introduction/pagination/index.mdx
```diff
@@ -17,22 +17,22 @@ To move between pages of a collection of records, use the links in the `next` an
 - ATTRIBUTE
   - DATA TYPE
   - DESCRIPTION
-- _links
+- \_links
```
jeesunikim: This shouldn't have a backtick

vcarl: Should this break anything? I think this was a prettier change, can revert it if we want to avoid introducing more characters.

src/api/resources/operations/object/set-options.mdx
```diff
@@ -0,0 +1,4 @@
+---
+title: Set Opitons object
```
vcarl:
```suggestion
title: Set Options object
```

src/api/introduction/pagination/index.mdx
jeesunikim: I don't think it will break anything. It might display differently though

PR #31, opened 2020-02-14
ApiRef-Polish

jeesunikim:
* Comments added in each file
* Renamed `DocPrismStyles` to `PrismStyles` since we no longer separate the two
* Renamed `NewDocText` to `Text` since we no longer separate the two


src/assets/icons/chevron.svg
```diff
@@ -1,6 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="9" height="10" viewBox="0 0 9 10">
     <g fill="none" fill-rule="evenodd">
-        <path fill="#FAFAFA" d="M-319-619H41v3948h-360z"/>
```
jeesunikim: Removed because it adds a background color

src/basics/Icons.js
```diff
@@ -2,31 +2,8 @@ import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 
-import { CSS_TRANSITION_SPEED } from "constants/styles";
-import { ARROW_SIZES } from "constants/arrows";
-
-import ArrowIcon from "assets/icons/arrow.svg";
 import CloseIcon from "assets/icons/close.svg";
 
-const ArrowEl = styled.div`
```
jeesunikim: Removed because it's an old style arrow. We got a new arrow style now.

src/basics/Inputs.js
```diff
@@ -2,10 +2,20 @@ import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 
-import { ARROW_SIZES } from "constants/arrows";
 import { FONT_WEIGHT, PALETTE } from "constants/styles";
 
-import { Arrow } from "basics/Icons";
+import Chevron from "assets/icons/chevron.svg";
```
jeesunikim: Replaced an old arrow with a new chevron

src/basics/Prism.js
```diff
@@ -0,0 +1,147 @@
+import { createGlobalStyle } from "styled-components";
+
```
jeesunikim: Renamed `NewDocPrism` to `Prism.js` since our new stellar.org launched

src/basics/Inputs.js
vcarl: It might be nice to keep this import as `import { Arrow } from 'basics/icons'` so we can abstract out the exact source of the icon? I'm not sure if we're using `Arrow` in multiple places though.

src/basics/Text.js
```diff
@@ -232,7 +248,7 @@ export const Bold = styled.strong`
 export const Muted = styled(Text).attrs(() => ({ as: "span" }))`
   color: ${({ theme }) => theme.medium};
 `;
-export const Strike = styled(Text).attrs(() => ({ as: "del" }))`
+export const Strike = styled(Text)`
```
vcarl: Do we want to remove the `del` tag? I think that helps e.g. screenreaders identify the content.

src/basics/Inputs.js
jeesunikim: I'll move `<CloseIcon/>` in `basics/icons` as well

PR #34, opened 2020-02-20
Add the transactions resource

kylemccollom:
- Missing the `post` endpoint

### DIFF
--- a/src/api/resources/transactions/list.mdx
+++ b/src/api/resources/transactions/list.mdx
@@ -1,0 +1,253 @@
+---
+title: List All Transactions
+order: 60
+---
+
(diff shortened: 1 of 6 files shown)


src/api/resources/transactions/list.mdx
```diff
@@ -0,0 +1,253 @@
...
+};
+
+var es = server.transactions()
+  .cursor('now')
+  .stream({
+      onmessage: txHandler
+  })
+```
+
+</CodeExample>
```
kylemccollom: Tacked on the streaming example at the end. Is that the correct placement? It should end up looking something like:
streaming design

PR #36, opened 2020-02-21
Add transactions

kylemccollom:
- Missing post

### DIFF
--- a/content/api/resources/transactions/list.mdx
+++ b/content/api/resources/transactions/list.mdx
@@ -1,0 +1,253 @@
+---
+title: List All Transactions
+order: 50
+---
+
+import { Endpoint } from "components/Endpoint";
+import { ExampleResponse } from "components/ExampleResponse";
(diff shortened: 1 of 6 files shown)


content/api/resources/transactions/list.mdx
```diff
@@ -0,0 +1,253 @@
...
+};
+
+var es = server.transactions()
+  .cursor('now')
+  .stream({
+      onmessage: txHandler
+  })
+```
+
+</CodeExample>
```
kylemccollom: Tacked on the streaming example at the end. Is that the correct placement? It should end up looking something like:
74986784-c2ec6d00-5407-11ea-934b-1711a5de21f2

jeesunikim: Yes. It looks like the JSON Example Response is missing padding at the bottom. Will add it in my next polish branch.

jeesunikim:
I'm investigating why `Transactions` isn't showing up under `Resources` in Side Nav

PR #37, opened 2020-02-21
feature/api_ref_new_structure > new structure for API Ref navigation to remove redundancy

jeesunikim:
**Before:**
api_ref_structure

**After:**
new_api_ref_structure

Updates: **1.** There are cases when folder's index.mdx shares the same title as its metadata.json's title. We are preventing redundancy by replacing its sub navigation's title to "Overview" if its title is the same as its metadata.json's. **2.** Nested navigation's default index.mdx has the same title as its folder's metadata.json's title which causes redundancy. For example, "Ledgers" navigation is created from metadata.json but its content that is in its index.mdx also has the title "Ledgers" which creates two "Ledgers" in the navigation. We are omitting redundancy by skipping renderItem() if that's the case

### DIFF
--- a/src/components/SideNav/SideNav.js
+++ b/src/components/SideNav/SideNav.js
@@ -66,7 +66,7 @@ export const SideNavBody = ({
   depth = -1,
 }) => (
   <NestedUl isOpen={isOpen}>
-    {items.map(({ id, title: subTitle, items: subItems }, index) => (
+    {items.map(({ id, title: subTitle, items: subItems, order }, index) => (
       // eslint-disable-next-line react/no-array-index-key
       <ListItemEl key={index}>
         <NestedNav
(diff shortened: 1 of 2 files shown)


src/components/SideNav/SideNav.js
```diff
@@ -111,9 +113,25 @@ const NestedNav = ({
   });
   const isOpen = isChildActive || isActive;
 
+  /* Nested navigation's default index.mdx has the same title as 
+    its folder's metadata.json's title which causes redundancy. 
+    For example, "Ledgers" navigation is created from metadata.json 
+    but its content that is in its index.mdx also has the title "Ledgers"
+    which creates two "Ledgers" in the navigation. We are omitting 
+    redundancy by skipping renderItem() if that's the case */
```
vcarl: Nice! Good catch

kylemccollom:
This looks great. There is a scenario in Aggregations where there is no index.mdx file, but an "Overview" page title appears in the navbar. Is there a work around for that?
Screen Shot 2020-02-21 at 1 41 49 PM
I could add an index.mdx in the future.

jeesunikim:
Fixed @kylemccollom

PR #39, opened 2020-02-24
bugfix/sidenav_no_relativepath > check to see if el.parent exists to avoid whiteouts

jeesunikim:
Adding an additional check to see if `el.parent` exists

### DIFF
--- a/src/components/SideNav/SideNav.js
+++ b/src/components/SideNav/SideNav.js
@@ -29,6 +29,10 @@ const ListItemEl = styled(ListItem)`
   margin: 0;
 `;
 
+/* Check to see if the parent directory has index.mdx */
+const checkIfOverview = (relativePath) =>
+  relativePath.split("/")[relativePath.split("/").length - 1] === "index.mdx";
+
 export const SideNav = ({ children, ...props }) => (
   <El {...props}>{children}</El>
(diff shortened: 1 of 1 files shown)


src/components/SideNav/SideNav.js
```diff
@@ -29,6 +29,10 @@ const ListItemEl = styled(ListItem)`
   margin: 0;
 `;
 
+/* Check to see if the parent directory has index.mdx */
+const checkIfOverview = (relativePath) =>
+  relativePath.split("/")[relativePath.split("/").length - 1] === "index.mdx";
```
vcarl: If this is getting broken out anyway, might be better to put `relativePath.split('/')` into a variable

src/components/SideNav/SideNav.js
```diff
@@ -129,12 +128,9 @@ const NestedNav = ({
               key={index}
               items={el.items}
               title={el.title}
-              relativePath={el.parent.relativePath}
+              relativePath={el.parent && el.parent.relativePath}
```
vcarl: Nit on naming: `el` to me sounds like a DOM node/ref, maybe `subnavItem` instead? Not something I'd block the PR on though.

PR #41, opened 2020-02-24
Polish/docs

jeesunikim:
- [Side Navigation > Hover effect activates when link isn't under mouse](https://app.asana.com/0/1119309091112078/1163196225281817)
- [make buttons/inputs same as on stellar.org](https://app.asana.com/0/1119309091112078/1162530625917539)
- [need hover states for side menu](https://app.asana.com/0/1119309091112078/1162905089549057)
- [When you click on a submenu item, the menu collapses, but should stay open with the page you're on as current](https://app.asana.com/0/1119309091112078/1163350873704090): 
-- The bug is that if the navigation has double nested items (for example, issuing assets got 7 items and its ‘connecting to outside infrastructure got its own nested items). In case like this, if you click a navigation item that does not have nested items, its active parent directory does not stay open.

- polish/docs > highlight active item in nav: Created a parent styled component because attaching props was throwing [Unknown Prop Errors](https://reactjs.org/warnings/unknown-prop.html)

### DIFF
--- a/src/components/Documentation/Articles.js
+++ b/src/components/Documentation/Articles.js
@@ -25,6 +25,10 @@ const ArticleLink = styled(Link)`
   color: ${(props) => (props.depth === 0 ? PALETTE.black80 : PALETTE.black60)};
   padding-left: ${(props) => (props.depth > 1 ? `${props.depth - 1}rem` : 0)};
   display: block;
+
+  &:hover {
+    color: #999;
+  }
 `;
 const ModifiedArrowIcon = styled(ArrowIcon)`
   position: absolute;
--- a/src/components/Documentation/Articles.js
+++ b/src/components/Documentation/Articles.js
@@ -24,7 +24,7 @@ const ArticleLink = styled(Link)`
   display: block;
 
   &:hover {
-    color: #999;
+    color: ${PALETTE.lightGrey};
   }
 `;
 const ModifiedArrowIcon = styled(ArrowIcon)`
--- a/src/components/Documentation/Articles.js
+++ b/src/components/Documentation/Articles.js
@@ -24,6 +24,7 @@ const ArticleLink = styled(Link)`
   font-size: ${(props) => (props.depth > 0 ? "0.875rem" : "1rem")};
   color: ${(props) => (props.depth === 0 ? PALETTE.black80 : PALETTE.black60)};
   padding-left: ${(props) => (props.depth > 1 ? `${props.depth - 1}rem` : 0)};
+  display: block;
 `;
 const ModifiedArrowIcon = styled(ArrowIcon)`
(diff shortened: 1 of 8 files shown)


src/components/Documentation/Articles.js
```diff
@@ -21,9 +21,11 @@ text-decoration: none;
 
 const ArticleLink = styled(Link)`
   ${topLevelNavItem}
-  font-size: ${(props) => (props.depth > 0 ? "0.875rem" : "1rem")};
-  color: ${(props) => (props.depth === 0 ? PALETTE.black80 : PALETTE.black60)};
-  padding-left: ${(props) => (props.depth > 1 ? `${props.depth - 1}rem` : 0)};
+  display: block;
+
+  &:hover {
+    color: #999;
```
piyalbasu: use `PALETTE.lightGrey` here?

src/helpers/documentation.js
```diff
@@ -72,16 +72,16 @@ export const findInitialOpenTopics = (data, pagePath, rootDir) => {
     initialTopicsState[relPath] = relPath === pgPath;
 
     const pathSegments = relPath.split("/");
+    const currentPathSegments = pgPath.split("/");
 
-    // if path is nested, open the parent dir as well
-    if (pathSegments.length > 2) {
-      const pagePathSegments = pgPath.split("/");
+    if (currentPathSegments.length > 2) {
```
jeesunikim: It was previously checking to see if all of the parent directories (`data`) instead of current active directory's path. For example, `issuing assets` got 7 items and I click non-nested navigation but because it got a `connecting to outside infrastructure` navigation which has its own nested items, it previously went through this if statement and prevented its navigation to stay open.

src/components/Documentation/Articles.js
```diff
@@ -72,20 +74,31 @@ const ArticleList = styled.ul`
...
-const Article = ({ article = {}, depth }) => {
+const CustomList = styled.li`
+  ${ArticleLink} {
+    font-size: ${(props) => (props.depth > 0 ? "0.875rem" : "1rem")};
+    color: ${(props) =>
+      props.depth === 0 ? PALETTE.black80 : PALETTE.black60};
+    padding-left: ${(props) => (props.depth > 1 ? `${props.depth - 1}rem` : 0)};
+    font-weight: ${(props) =>
+      props.isActive ? FONT_WEIGHT.bold : FONT_WEIGHT.normal};
+  }
```
piyalbasu: we don't have to do this now, but at some pt we should figure out how to combine this with the styles for `NavItemEl` in /templates/ApiReference

jeesunikim: Good point. I can put it in `SharedStyle` folder tomorrow

PR #45, opened 2020-02-26
Polish/docs > Fixing its 'NEXT UP' link

jeesunikim:
- [/issuing-assets/control-asset-access/'s next up doesn't link to its next one](https://app.asana.com/0/1130719348309291/1163413894876105)

issuing_assets

For example, `Issuing Assets > Control Asset Access` was pointing at its next parent level `Run an API Server`'s `Administration` instead of its sibling element `Connecting To Outside Infrastructure` because its `nextUp()` that links to its next element is based on `metadata.json` order.

See the following:
sortedDocs

Because the next element after `docs/issuing-assets` is `docs/run-core-node/administration`, `Issuing Assets > Control Asset Access` was pointing at `run-core-node/administration` instead of its nested element.

So I corrected the order of metadata.json but let me know if I should use a sorting algorithm to re-sort `sortedDocs`.

- Added the missing `<CodeExample/>` so I won't get errors in my console and it looks good

### DIFF
--- a/src/basics/Text.js
+++ b/src/basics/Text.js
@@ -222,7 +222,7 @@ export const Preformatted = styled.pre`
     line-height: inherit;
     white-space: pre-wrap;
     max-width: 90rem;
-    width: 1000rem;
+    width: 100%;
     word-break: break-all;
   }
   && .line-numbers-rows {
--- a/src/basics/Text.js
+++ b/src/basics/Text.js
@@ -222,7 +222,7 @@ export const Preformatted = styled.pre`
     line-height: inherit;
     white-space: pre-wrap;
     max-width: 90rem;
-    width: 100%;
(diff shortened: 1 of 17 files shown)


src/basics/Text.js
```diff
@@ -222,7 +222,7 @@ export const Preformatted = styled.pre`
     line-height: inherit;
     white-space: pre-wrap;
     max-width: 90rem;
-    width: 1000rem;
+    width: 100%;
```
vcarl: We need to force this to something super wide in order to avoid wrapping URLs. I noticed that it's always at 90rem though, so might be be better to change to `width: 90rem;`

jeesunikim: Removed it because the `<Text />` that docs was using was a placeholder

PR #48, opened 2020-02-27
fix next up link to work with nested articles

piyalbasu:
My original implementation didn't take nested articles into account. whoops!

### DIFF
--- a/src/helpers/documentation.js
+++ b/src/helpers/documentation.js
@@ -31,31 +31,6 @@ export const query = graphql`
 export const buildRelPath = (relativeDirectory, rootDir) =>
   relativeDirectory.replace(rootDir, "") || "/";
 
-export const nextUp = (topicArr, topicIndex, childArr, childIndex) => {
-  // End of list
-  if (topicIndex + 1 === topicArr.length) {
-    return { title: "", url: "" };
-  }
-
-  // Go to next topic
-  if (childIndex === childArr.length - 1) {
-    const nextTopic = topicArr[topicIndex + 1].nodes[0];
-    return {
-      title: nextTopic.fields
-        ? nextTopic.fields.metadata.data.title
-        : "MISSING METADATA.JSON",
-      url: `${buildPathFromFile(nextTopic.relativePath)}`,
-    };
-  }
-
-  // Go to next article in topic
-  const nextChild = childArr[childIndex + 1];
-  return {
-    title: nextChild.childMdx.frontmatter.title,
-    url: buildPathFromFile(nextChild.relativePath),
-  };
-};
-
 /**
  * findInitialOpenTopics builds an object of booleans signifying which nav items should be in an open state on page load.
  * @param {object} data Raw data from graphQL query.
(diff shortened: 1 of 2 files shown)


src/helpers/documentation.js
```diff
@@ -235,9 +209,85 @@ export const buildDocsContents = (data, rootDir) => {
     insertPageData(relPath, contents, articles, rootPageData);
   });
 
+  // Now that content is in proper order, travel through tree and add Next Up link to each article
```
piyalbasu: my mistake before was adding the next up links before we properly grouped all the data. That's why things were in the right order in the left nav, but the next up links for nested items did not work at all

src/helpers/documentation.js
```diff
@@ -235,9 +209,85 @@ export const buildDocsContents = (data, rootDir) => {
...
 };
 
+/**
+ * addNextUpToArticles recursively travels through contents tree to add a Next Up property to each
+ * @param {array} articles List of articles of the current topic.
+ * @param {number} articleIndex The index of the article we're working with
+ * @param {number} nextTopic List of articles of the next topic
+ * @returns {void}
+ */
+const addNextUpToArticles = (articles, articleIndex, nextTopic) => {
```
piyalbasu: take each section of articles, and go through each one and add the next up link to each one.

if we encounter a nested group of articles, run through this func with that group

src/helpers/documentation.js
```diff
@@ -235,9 +209,85 @@ export const buildDocsContents = (data, rootDir) => {
...
+};
+
+/**
+ * nextUpLink returns the url and title of the next article, whether it's in the same topic or if it's in the next topic
+ * @param {array} topicArticles List of articles of the current topic.
+ * @param {number} articleIndex The index of the article we're working with
+ * @param {number} nextTopic List of articles of the next topic
+ * @returns {object} { title, url }
+ */
+export const nextUpLink = (topicArticles, articleIndex, nextTopic) => {
```
piyalbasu: figure out if the next article is a sibling or if the next article is the first child of the next topic

src/helpers/documentation.js
```diff
@@ -235,9 +209,85 @@ export const buildDocsContents = (data, rootDir) => {
...
+    url: nextNestedFirstArticle.url,
+  };
+};
+
+/**
+ * findNextFirstArticle recursively drills down nested articles to find the first article we can link to
+ * @param {object} articleData List of articles of the current topic.
+ * @returns {object} { title, url }
+ */
+const findNextFirstArticle = (articleData) => {
```
piyalbasu: this helper method just find the next article we can link to. in some cases, it's just the next sibling. but in others, it's a nested collection and we have to drill down to find an article

jeesunikim:
Awesome! Thanks @piyalbasu

vcarl:
Now that we have Jest set up, it'd be dope to get some unit tests in place for this @piyalbasu. Even if it's just a matter of extracting some fixture data and ensuring it transforms correctly, that'd be a nice boost to confidence if we have to touch this in the future.

PR #59, opened 2020-03-09
Add dockerfile for hosting on k8s

vcarl:


nginx/includes/server.conf
```diff
@@ -0,0 +1,18 @@
...
+# from each individual section.
+set $no-js false;
+if ($arg_javascript = 'false') {
+  set $no-js true;
+}
+if ($no-js = false) {
+  rewrite ^/api/.*$ /api/;
+}
+if ($no-js = true) {
+  rewrite ^/api/(.*) /no-js/api/$1;
```
piyalbasu: 👍

Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM ubuntu:16.04 as build
+
+MAINTAINER SDF Ops Team <ops@stellar.org>
```
jacekn: I think it should be your team rather than Ops? Not a big deal tough, can be fixed later.

README.md
```diff
@@ -17,23 +18,32 @@ Each folder must have a `metadata.json` file with 3 keys:
...
+
+## Local production build
+
+The build has been dockerized so we can host with nginx on Kubernetes.
+
+To build and run locally:
+
+```sh
+docker image build -t new-docs:0.0 .
+docker run -p 8000:80 new-docs:0.0
```
jacekn: Maybe use more explicit tag, for example `new-docs:localbuild` ?

jenkins.bash
```diff
@@ -0,0 +1,6 @@
+#! /usr/bin/env bash
+set -e
+
+yarn install
+
+yarn build
```
jacekn: Where is this file used?

nginx/nginx.conf
```diff
@@ -0,0 +1,21 @@
+events { }
+http {
+  include mime.types;
+  server {
+    server_name docs.stellar.org;
```
jacekn: It would be good if the nginx accepted any server_name. Reason for that is that we could then use the same image for any dev or PR-preview style deployment without need to modify it.

jenkins.bash
vcarl: Currently nowhere, I was anticipating this being used in Jenkins once we have a job set up though.

jacekn: It's easier to build using `docker build`.

FYI a few other projects we also added Makefile for convenience. For example: https://github.com/stellar/dashboard/blob/master/Makefile
It could be useful for development, for example you could add something like `make localrun` target to build and run the image locally.

Dockerfile
vcarl: Cool that sounds good to me. I used the [stellar.org-config](https://github.com/stellar/stellar.org-config/blob/master/nginx/nginx.conf) config as a starting point, I left this alone. I can put my email.

jenkins.bash
vcarl: Good point. We'll probably use npm scripts to wrap it, make is unlikely to be more familiar for folks touching this code.

nginx/nginx.conf
vcarl: I copied the format from [stellar.org-config](https://github.com/stellar/stellar.org-config/blob/master/nginx/nginx.conf), will need some guidance on how to change it if this doesn't work.

jacekn: Ah interesting. So it works because the first vhost is also the default one ([docs](https://nginx.org/en/docs/http/request_processing.html)).
So it's fine to merge as is but if you want to be explicit you could do something like this:
```
server {
    listen      80 default_server;
    server_name docs.stellar.org;
}
```

PR #62, opened 2020-03-11
Polish fonts, metadata, consolidate layouts

vcarl:
We had some cruft in our style constants and layouts because of how the codebase has evolved, so I dug through and removed some of the complexity. `DocsBase` isn't relevant now that there's only one type of layout.

* Remove fonts that were 404ing
* Switch some `block`ing fonts to `swap`, add some links to what character sets are in which fonts
* Remove some unused constants
* Add descriptions
* Optimize Documentation query

I realized that we were fetching the body of every doc page when building each doc page, so our query was returning N results of which we used 1. Since I had to add a second form of the body to the query, I didn't want to fetch 2N results when we'd use 2. This change appears to be a significant build speed improvement, and should help prevent some pain as we add more content.
before:
```
$ yarn build
…
✨  Done in 61.91s.
```
after:
```
$ yarn build
…
✨  Done in 31.32s.
```


jest.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     ".+\\.(css|styl|less|sass|scss)$": `identity-obj-proxy`,
     ".+\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": `<rootDir>/__mocks__/file-mock.js`,
   },
-  testPathIgnorePatterns: [`node_modules`, `.cache`, `public`],
+  testPathIgnorePatterns: [`node_modules`, `.cache`, `public`, `fixtures`],
```
vcarl: cc @jeesunikim, this lets us add `__tests__/fixtures/` folders, which I did for an MDX helper below.

PR #65, opened 2020-03-12
Polish/docs api

jeesunikim:
[Implement only one language available on <CodeExample/>](https://app.asana.com/0/1119309091112078/1166260941039554)
[Align line numbers with code title](https://app.asana.com/0/1119309091112078/1165136694034025)
[need better state change on side menu](https://app.asana.com/0/1119309091112078/1162901872838449)
[Add line spacing to the "NavTable" component](https://app.asana.com/0/1119309091112078/1163771225313413)

`<PrismStyles/>` has been moved from a component level to a template level because having it in component level would allow overriding styles if it styling was based on props. For example, previously we had added props to `<CodeExample/>` component to have different look than `<ExampleResponse/>` for json but because it was done in component level, its global style created by `<CodeExample/>` would override `<ExampleResponse/>`.override each other.


content/docs/building-apps/connect-to-anchors/deposit-anchored-assets.mdx
```diff
@@ -188,18 +187,14 @@ That selection will come back as a string in the form of `{code}:{issuer}` so we
 <CodeExample>
 
 ```ts
-...
```
jeesunikim: @kylemccollom I noticed redundant **```**. It doesn't break anything. Just an FYI - let's go through code snippets before we launch!

content/docs/building-apps/connect-to-anchors/deposit-anchored-assets.mdx
```diff
@@ -259,33 +246,29 @@ The endpoint for this JWT endpoint is noted by the `WEB_AUTH_ENDPOINT` on the TO
...
 ```
 
 </CodeExample>
 
 This call is our first little bit of magic. Once we setup our `multipart/form-data` entry values we can POST them to the `${this.toml.TRANSFER_SERVER}/transactions/withdraw/interactive` endpoint along with our `auth` Authorization Bearer `auth` token we acquired earlier. The response to this call will contain a special url for interacting directly with the anchor providing the data they need to accept our deposit asset.
 
-<CodeExample>
+<ExampleResponse title="Example Response">
 
 ```json
```
jeesunikim: Just to confirm - we should use <ExampleResponse/> for json format right? @kylemccollom

kylemccollom: let's try it and see. that's more for api reference, but could work well in docs, too.

PR #68, opened 2020-03-17
Use folder title for `index.mdx` docs

vcarl:
before:
Screen Shot 2020-03-17 at 2 12 19 PM
after:
Screen Shot 2020-03-17 at 2 12 27 PM

### DIFF
--- a/src/components/Documentation/Articles.js
+++ b/src/components/Documentation/Articles.js
@@ -91,12 +91,11 @@ const CustomList = styled.li`
   }
 `;
 
-const Article = ({ article = {}, activeItem, depth }) => {
-  const { id, title, url } = article;
+const Article = ({ title, url, activeItem, depth }) => {
   const isActive = url.includes(activeItem);
 
   return (
-    <CustomList key={id} isActive={isActive} depth={depth}>
+    <CustomList isActive={isActive} depth={depth}>
       <ArticleLink href={url}>{title}</ArticleLink>
     </CustomList>
   );
@@ -105,11 +104,12 @@ const Article = ({ article = {}, activeItem, depth }) => {
 Article.propTypes = {
   depth: PropTypes.number,
   activeItem: PropTypes.string,
+  title: PropTypes.string,
+  url: PropTypes.string,
 };
 
 const Articles = ({
   articles,
-  id,
   initialTopicsState,
   isNested,
   title,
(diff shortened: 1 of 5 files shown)


src/components/Documentation/Articles.js
```diff
@@ -126,12 +140,27 @@ const Articles = ({
...
+
   return (
-    <li key={id}>
+    <El>
+      {/* eslint-disable-next-line */}
       {isNested ? (
-        <NestedArticleTopicExpander onClick={() => topicToggleHandler()}>
-          {title}
-        </NestedArticleTopicExpander>
+        indexArticle ? (
```
vcarl: Sorry, this is gross, but I didn't want to pull it off into a variable. Can revisit if you'd rather not see a nested ternary (which I get)

src/components/Documentation/Articles.js
```diff
@@ -144,46 +173,41 @@ const Articles = ({
...
-            />
-          </ArticleList>
-        ),
-      )}
-    </li>
+      <ArticleList isCollapsed={isCollapsed}>
+        {Object.entries(articles)
+          .filter(([filename]) => filename !== "index")
+          // Only get second arg
+          .map(([, article]) =>
```
vcarl: I low key forgot you could do this. I usually do `([_, thing])`, which I prefer because it signifies that there's a value there that's not being used, but the eslint "no unused" doesn't have any exceptions.

src/helpers/__tests__/documentation.js
```diff
@@ -0,0 +1,8 @@
+import { buildDocsContents } from "../documentation";
+import { input, output } from "./fixtures/documentation";
+
+describe("buildDocsContents", () => {
+  it("matches snapshot", () => {
+    expect(buildDocsContents(input.basic, "docs")).toStrictEqual(output.basic);
```
vcarl: Suuuper basic manual snapshot test. Basically just here so that if we have to touch this function again, can at least check happy path.

jeesunikim:
pr68
Are titles - `Make XLM Payments` or `Handle Custom Assets` - being repeated in submenu intentional? @vcarl

src/helpers/__tests__/fixtures/documentation.js
```diff
@@ -0,0 +1,430 @@
+export const input = {
```
jeesunikim: nice!

src/helpers/__tests__/documentation.js
jeesunikim: I like this!

piyalbasu: this is great! i think something simple like this is a really good idea

vcarl: Cheers. My main complaint with snapshot testing is that it's super easy to become numbed to test failures. For asserting complex output doesn't change (like here) I think I like it.

vcarl:
@jeesunikim That looks like a content issue, so I'm gonna leave it alone for now. There's an empty `index.mdx` under "Make XLM Payments" and a separate `xlm-payments.mdx` with the same title

PR #77, opened 2020-03-24
bugfix/docs-order > order its parent at the end instead of in the beginning

jeesunikim:
Instead of sorting its parent `metadata.json`'s order in the beginning, I sorted it in the end. 

For example, previously, `/building-apps/create-stellar-wallet/index.mdx` was placed before `/building-apps/index.mdx` because the order number of `metadata.json` of `/building-apps/create-stellar-wallet` is less than the order number of `metadata.json` of `/building-apps/`. We start out with correctly ordered mdx files thanks to `DocumentationQuery` but running sorting parent function in the beginning ignored its metadata.json's hierarchy. 

Before (by sorting parent metadata in the beginning, its order is incorrect)
before

Correct order done by graphql
after

By sorting parent metadata in the end, we keep the children's order in correct place and just sort the parent
final

### DIFF
--- a/src/helpers/documentation.js
+++ b/src/helpers/documentation.js
@@ -2,7 +2,7 @@ import { graphql } from "gatsby";
 
 import { groupBy } from "helpers/groupBy";
 import { buildPathFromFile } from "helpers/routes";
-import { compareOrders } from "helpers/sortReference";
+import { compareNestedEntries, compareOrders } from "helpers/sortReference";
 
 export const DOCS_CONTENT_URL =
   "https://github.com/stellar/new-docs/blob/master/content/";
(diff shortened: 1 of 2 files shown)


src/helpers/sortReference.js
```diff
@@ -1,5 +1,5 @@
 export const compareOrders = (a, b) => a.order - b.order;
-const compareNestedEntries = (a, b) => compareOrders(a[1], b[1]);
+export const compareNestedEntries = (a, b) => compareOrders(a[1], b[1]);
```
jeesunikim: These two functions are so useful @vcarl

src/helpers/documentation.js
```diff
@@ -220,7 +214,15 @@ export const buildDocsContents = (data, rootDir) => {
...
     addNextUpToArticles(topicArticles, 0, nextTopic);
   });
-  return contents;
+
+  /* After its nested page data are sorted and added in,
+  sort its parent by topicOrder which is metadata.data.order */
+  const sortedDocs = Object.entries(contents)
+    .sort(compareNestedEntries)
+    /* reverse it to object */
+    .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {});
```
vcarl: There's also [`Object.fromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries)

(oops I see your commit, bummer that this isn't supported in node yet)

PR #78, opened 2020-03-25
Bugfix/docs

jeesunikim:
- [When there is no number placed for order, do alphabetical order](https://app.asana.com/0/1119309091112078/1168028367950529)
- [Fix NextUp link](https://app.asana.com/0/1119309091112078/1168252372807489)
- Texts to bold when active for `Overview` and `IndexArticle`

### DIFF
--- a/src/helpers/documentation.js
+++ b/src/helpers/documentation.js
@@ -28,6 +28,7 @@ export const query = graphql`
             data {
               order
               title
+              sortMethod
             }
           }
         }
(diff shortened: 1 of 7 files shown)


src/helpers/documentation.js
```diff
@@ -181,6 +185,13 @@ export const buildDocsContents = (data, rootDir) => {
       : "MISSING METADATA.JSON";
     const articles = {};
 
+    /* if the nodes are coming from /glossary
+    which doesn't have an order specificed in .mdx file
+    sort the nodes alphabetically */
+    if (topic.nodes[0].relativePath.includes("docs/glossary/")) {
```
vcarl: Maybe it'd be feasible to change this to be a folder metadata flag, like `"sortMethod": "alphabetical"` in metadata.json? We've got a number of hardcoded paths already and I'm not really a fan of it (despite writing several of them). I don't think this _needs_ to be changed though.

jeesunikim: I like that!

jeesunikim: Updated @vcarl

PR #83, opened 2020-03-30
Make API ref attribute list more resilient to content formatting

vcarl:
Major changes:

* Rename `NestedTable` -> `AttributeTable`—figured while I was overhauling, I'd take the chance to rename
* Add a helper method (with simple tests) to build up a tree of attributes, separating that from the process of rendering.

Reason for the change is that previously, the logic was relying on where in the hierarchy the elements were to render. This ended up breaking because MDX creates a more complex tree when there's formatting—elements that should have been descriptions with additional formatting were rendering as sub-attributes. This change walks the tree and looks for `li` elements (as described by the props MDX puts on elements). There's a risk that this could break badly in the future, if MDX changes how it addresses components. That seems relatively unlikely though.

I made a small design adjustment to simplify implementation, also

before:
stellar-documentation netlify com_no-js_api_resources_accounts_object_
after:
localhost_8000_no-js_api_resources_accounts_object_


vcarl:
Sorry for file noise! Major changes are [AttributesTable.js](https://github.com/stellar/new-docs/pull/83/files?file-filters%5B%5D=#diff-ca09bfe30c807a3406265c8f1637df78) and [documentation.js](https://github.com/stellar/new-docs/pull/83/files?file-filters%5B%5D=#diff-23ebbc9236fe8355925b895c936ef6d1)

jeesunikim:
So much cleaner! Love it.

PR #85, opened 2020-03-31
Bugfix/scroll > allowing independent scrolling panels with no body scrolls

jeesunikim:
[Fix sidebar scroll](https://app.asana.com/0/1119309091112078/1163228700401441)
- When you scroll to the bottom of the API ref or doc' sidebar nav, scrolling should stop and not affect the content column. Right now, when you scroll to the bottom of the sidebar nav and keep scrolling, it causes the content column to scroll down as well.

- Disabling body scroll means `window.addEventListener("scroll")` no longer works since its content scroll is based on the `<Column/>` instead of `window` in `ScrollRouter.js``SideNav/Provider.js`
- I thought about using `useRef` instead of `document.querySelector`, but am not sure about passing the ref around. Let me know.

### DIFF
--- a/src/components/SideNav/Provider.js
+++ b/src/components/SideNav/Provider.js
@@ -1,6 +1,8 @@
 import React from "react";
 import PropTypes from "prop-types";
 
+import { DOM_TARGETS } from "constants/domNodes";
+
 import { findActiveNode } from "helpers/dom";
 
 const sortByPosition = (a, b) => {
@@ -24,7 +26,7 @@ export const Provider = ({ children }) => {
   );
 
   React.useEffect(() => {
-    const contentDom = document.querySelector("#content-column");
+    const contentDom = document.querySelector(`#${DOM_TARGETS.contentColumn}`);
 
     const handler = () => {
       // If we haven't scrolled at least 20 pixels, just bail.
--- a/src/components/SideNav/Provider.js
+++ b/src/components/SideNav/Provider.js
@@ -24,9 +24,11 @@ export const Provider = ({ children }) => {
   );
 
   React.useEffect(() => {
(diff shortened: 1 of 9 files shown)


src/components/SideNav/Provider.js
```diff
@@ -24,28 +24,31 @@ export const Provider = ({ children }) => {
   );
 
   React.useEffect(() => {
+    const contentDom = document.querySelector("#content-column");
```
jeesunikim: I thought about using `useRef` in `Documentation.js` or `ApiReference.js` instead of `document.querySelector`, but am not sure about passing the ref around. Let me know.

vcarl: I think this is solid. There have been several spots now where, after getting it working, I realized I probably could have done some query selectors and simplified the end result, passing refs around has a lot of weird plumbing.

vcarl: There's a `constants/domNodes.js` (currently just has portal targets, which are mostly unused), might be best to add this id as a constant in there

src/components/SideNav/SideNav.js
```diff
@@ -128,7 +128,7 @@ const NestedNav = ({
               key={index}
               items={subnavItem.items}
               title={subnavItem.title}
-              relativePath={subnavItem.parent && subnavItem.parent.relativePath}
+              relativePath={subnavItem.parent?.relativePath}
```
vcarl: 🎉

src/components/SideNav/Provider.js
jeesunikim: Updated!

vcarl: Looks good!

PR #92, opened 2020-04-01
Bugfix/docs-api

jeesunikim:
[Allow checkmarks](https://app.asana.com/0/1119309091112078/1169423234140093)
* New checkmark icon added

[Page Outline - Click and Scroll is not working](https://app.asana.com/0/1119309091112078/1169423234140097)

[H4s following H3s don't have a line break](https://app.asana.com/0/1119309091112078/1169423234140092)

### DIFF
--- a/src/constants/docsComponentMapping.js
+++ b/src/constants/docsComponentMapping.js
@@ -63,7 +63,7 @@ export const components = {
       }
     }
   `,
-  h3: styled(TextComponents.LinkedH3)`
+  h3: styled(TextComponents.H3)`
     a {
       color: ${({ theme }) => theme.text};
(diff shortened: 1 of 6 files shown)


src/constants/docsComponentMapping.js
```diff
@@ -63,7 +63,7 @@ export const components = {
       }
     }
   `,
-  h3: styled(TextComponents.LinkedH3)`
+  h3: styled(TextComponents.H3)`
```
jeesunikim, line 66: `<LinkedH../>` elements got `display: inline-block` which caused it it not break. Since headers after H2 do not get linked, I removed it.

PR #95, opened 2020-04-03
Bugfix/api-mobile > api ref to be functional on mobile

jeesunikim:
[Disable Scroll Highlight functionality on mobile](https://app.asana.com/0/1119309091112078/1169602362766355)
- Disable the highlight as you scroll functionality on mobile
- Keep the "current" highlighting functionality on mobile
- Nested navigation items should be visible by default

* Using `document.documentElement.style.setProperty` so that our `height: 100vh;` will be always 100vh via its `window.innerHeight` on mobile devices when users have to zoom out to see the whole page since `/api` is using its desktop design on mobile
-- Based on [the trick to viewport units on mobile](https://css-tricks.com/the-trick-to-viewport-units-on-mobile/)

* Hide **Page Outlines** on the right column on `/docs` when there is no items available to be highlighted

**iphone-6-landscape**

iphone-6-landscape

**iphone-6-portrait**

iphone-6-portrait

### DIFF
--- a/src/basics/Grid.js
+++ b/src/basics/Grid.js
@@ -113,6 +113,8 @@ const ColumnEl = styled.div`
     css`
       overflow-y: scroll;
       height: 100vh;
+      /* --vh will be set via setViewportHeight() ApiReference.js */
+      height: calc(var(--vh, 1vh) * 100);
     `}
 `;
 
--- a/src/components/Navigation/SharedStyles.js
+++ b/src/components/Navigation/SharedStyles.js
@@ -22,6 +22,11 @@ export const NavAbsoluteEl = styled.div`
   top: 4rem;
   bottom: 4.31rem;
 
+  /* <SideNav/> does not have mobile friendly design (API REF) */
+  @media (${MEDIA_QUERIES.ltLaptop}) {
+    overflow-y: scroll;
+  }
+
   &:hover {
     overflow-y: scroll;
   }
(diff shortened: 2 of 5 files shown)


src/basics/Grid.js
```diff
@@ -113,6 +113,8 @@ const ColumnEl = styled.div`
     css`
       overflow-y: scroll;
       height: 100vh;
```
piyalbasu, line 115: looks like we don't need this old height property anymore, right?

src/components/Navigation/SharedStyles.js
```diff
@@ -140,6 +145,8 @@ export const NavItem = styled.div`
 export const StickyEl = styled.div`
   width: 100%;
   height: 100vh;
```
piyalbasu, line 147: same

src/basics/Grid.js
jeesunikim: I'd like to keep it just in case users visit it via browsers that don't support custom vh

src/templates/ApiReference.js
```diff
@@ -301,6 +304,29 @@ const ApiReference = React.memo(function ApiReference({ data, pageContext }) {
...
+      const vh = window.innerHeight * 0.01;
+      // Sets the value in the --vh custom property to the root of the document
+      document.documentElement.style.setProperty("--vh", `${vh}px`);
+    });
+
+    setViewportHeight();
+    window.addEventListener("resize", setViewportHeight);
+
+    return () => {
+      window.removeEventListener("scroll", setViewportHeight);
```
piyalbasu, line 326: why is the `removeEventListener` on `scroll`? Maybe i'm missing something, but wouldn't the `resize` listener remain after unmount?

jeesunikim: Good call

src/basics/Grid.js
piyalbasu: ah gotcha. that makes sense

PR #98, opened 2020-04-08
Mobile friendly docs pt 1

vcarl:
This is a lot of prep work to adjust layouts to something that can be made responsive/otherwise work well on mobile. There's still more work to be done on changing behaviors etc, but this has a solid number of changes going on already and I want to get it in.

~~I made a design adjustment to the overflow shadows on the nav for technical reasons:~~ Going to undo this and hide the fade effect when it starts behaving badly.
Screen Shot 2020-04-08 at 2 14 39 PM

The way it was set up before was doing weird stuff on oddball screen sizes, and with a different method that doesn't hit those issues, I wasn't able to replicate exactly the same effect (the shadow is _under_ the text now, so using the background color was invisible).

Also fixes some weird scrollbars on Windows. In part 2, I'll see about making the center column's scrollbar a little nicer, I think it's crowding the content atm. Before:
image
After:
image


src/components/Navigation/SharedStyles.js
```diff
@@ -14,60 +14,44 @@ export const H2 = styled(BasicH2)`
...
+  background: linear-gradient(${BACKDROP_COLOR} 30%, ${BACKDROP_FADE}),
+    linear-gradient(${BACKDROP_FADE}, ${BACKDROP_COLOR} 70%) 0 100%,
+    // Shadows
+      linear-gradient(${SHADOW_COLOR}, ${SHADOW_FADE} 100%),
+    linear-gradient(${SHADOW_FADE}, ${SHADOW_COLOR}) 0 100%;
+  background-repeat: no-repeat;
+  background-size: 100% 3rem, 100% 3rem, 100% ${SHADOW_SIZE},
+    100% ${SHADOW_SIZE};
+  // Opera doesn't support this in the shorthand
+  background-attachment: local, local, scroll, scroll;
```
vcarl, line 54: New scroll shadow method, stolen from the linked blog post above.

src/components/Navigation/SharedStyles.js
```diff
@@ -143,12 +125,7 @@ export const NavItem = styled.div`
...
 export const StickyEl = styled.div`
-  width: 100%;
   height: 100vh;
-  /* --vh will be set via setViewportHeight() ApiReference.js */
-  height: calc(var(--vh, 1vh) * 100);
-  overflow-y: scroll;
-  position: sticky;
-  top: 0;
-  z-index: 3;
+  display: flex;
```
vcarl, line 129: I changed out how this element's layout is done from absolute to flexbox, think it's a little more stable now

src/templates/ApiReference.js
```diff
@@ -342,6 +319,7 @@ const ApiReference = React.memo(function ApiReference({ data, pageContext }) {
           title="Stellar API Reference"
           description="The complete API reference for the Stellar network. Includes descriptions of Horizon endpoints, network concepts, and example code for some languages."
           pageContext={pageContext}
+          viewport="width=1366, initial-scale=.1"
```
vcarl, line 322: I had to do `initial-scale=.1` to get it to be zoomed out when the page loads.

src/templates/ApiReference.js
```diff
@@ -305,28 +304,6 @@ const ApiReference = React.memo(function ApiReference({ data, pageContext }) {
...
-      document.documentElement.style.setProperty("--vh", `${vh}px`);
-    });
-
-    setViewportHeight();
-    window.addEventListener("resize", setViewportHeight);
-
-    return () => {
-      window.removeEventListener("resize", setViewportHeight);
-    };
-  }, []);
```
piyalbasu, line 328: is this removal on purpose? didn't @jeesunikim  just add this?

jeesunikim: `viewport="width=1366, initial-scale=.1"` did the trick so it is no longer needed

jeesunikim:
@vcarl, I think the faded text background image should be confirmed by Charles.

src/basics/Grid.js
```diff
@@ -111,10 +111,10 @@ const ColumnEl = styled.div`
   ${({ isIndependentScroll }) =>
     isIndependentScroll &&
     css`
-      overflow-y: scroll;
-      height: 100vh;
-      /* --vh will be set via setViewportHeight() ApiReference.js */
-      height: calc(var(--vh, 1vh) * 100);
+      @media (${MEDIA_QUERIES.gtTablet}) {
```
jeesunikim, line 114: This needs to be enabled in mobile for Api Reference to be able to scroll the side navigation

src/components/Navigation/SharedStyles.js
jeesunikim: Let's confirm the design of this with Charles

src/templates/ApiReference.js
vcarl: It is on purpose, we chatted and it looks like the viewport settings solve the same problem

src/basics/Grid.js
vcarl: The viewport setting change makes it so the sidenav takes up the entire left side of the page, I haven't needed to scroll. Page looks like this for me
Screen Shot 2020-04-08 at 3 38 29 PM

jeesunikim: Gotcha! One thing, all the sub items are supposed to be shown on mobile by default. But I just checked this site on my phone and its scroll and highlight functionality work pretty fine so let me confirm with Charles on whether we should display all subnav by default.

src/components/Navigation/SharedStyles.js
vcarl: Instead of this, I'm gonna change it to hide the fade on small screens. Will circle back in my pt 2 PR.

src/basics/Grid.js
vcarl: Oh, I remember why I had to change this now. This style still applies on API Ref on mobile because the viewport is set to be wide, but having a scrollable div on mobile made /docs behave super weirdly. Was easy to get halfway past the nav, then start scrolling in the docs container while only half of it is onscreen.

jeesunikim: No need to display all nested items by default. All good 👍! Thank you.

PR #105, opened 2020-04-09
bugfix/grid > implemented a small hack in order to make grid work on a large screen

jeesunikim:
[Padding between columns](https://app.asana.com/0/1119309091112078/1162530625941540)
* This bug is for xl screen `(min-width: 1441px)`
* As Carl noted, it's impossible to set a minimum width for `column-gap`
* I implemented a hack to skip the first column for each column in the middle two sections on a xl screen to give 4rem padding

**Before**
before

**After**
after

### DIFF
--- a/src/templates/ApiReference.js
+++ b/src/templates/ApiReference.js
@@ -47,9 +47,9 @@ import { Route, SectionPathContext } from "components/ApiRefRouting/Route";
 import {
   ApiReferenceRow,
   ApiReferenceWrapper,
-  Container,
   NestedRow,
   SideNavColumn,
+  CustomColumn,
 } from "components/Documentation/SharedStyles";
 import { Expansion } from "components/Expansion";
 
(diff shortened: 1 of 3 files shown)


src/templates/ApiReference.js
```diff
@@ -232,7 +232,15 @@ const headerOptions = {
   treatIdAsHref: true,
   LinkComponent: DocsLink,
 };
-const ApiRefH1 = makeLinkedHeader(H1, headerOptions);
+/* Hack to make it look appear as if we had a column-gap
+4rem between colums and its h1 on a large screen */
+const XlH1 = styled(H1)`
+  @media (${MEDIA_QUERIES.gtXlDesktop}) {
+    margin-left: 4rem;
```
vcarl, line 239: I notice the gap gets misaligned at some screen sizes:
Screen Shot 2020-04-09 at 4 47 37 PM
Maybe instead of a fixed margin, the H1 in `ApiReference.js` should move within the grid? Might mean adding a new row though.

jeesunikim: Good call!

jeesunikim: Added a new row to surround `<H1/>` and removed `<Container/>` from ApiReference since ApiReference shouldn't have `max-width` which was coming form `<Container/>` previously.

PR #107, opened 2020-04-12
add the effects resource

kylemccollom:


kylemccollom:
can we add an h4 styling from charles to the "Account Effects", "Signer Effects", etc titles on the Effect Types page?

PR #111, opened 2020-04-13
Update scroll shadows, add beta badge

vcarl:
I swapped the header position to use `margin-scroll-top` because otherwise the page title covers the "open issues here" link. I think a slightly degraded experience is okay until straggler browsers update to support this. https://caniuse.com/#feat=mdn-css_properties_scroll-margin-top


src/components/Navigation/NavLogo.js
```diff
@@ -27,21 +27,29 @@ const PageNameEl = styled.span`
 `;
 const NavDividerEl = styled(DividerEl)`
   height: 3rem;
-  background-color: ${({ theme }) => theme.border};
+  background-color: ${PALETTE.white80};
```
jeesunikim, line 30: Is this new design? No more visible divider line?

vcarl: Hmm it was looking invisible to me beforehand. Lemme double check this, I may have committed too quickly.

jeesunikim: I think you are right (I asked Charles just in case since this wasn't brought up when we had in person QA).

vcarl: I changed this to `white60`, `white80` is the same as the background. Looks like `theme.border` is no more, which would explain why it's not showing up at https://developers.stellar.org/docs/ right now

PR #112, opened 2020-04-15
Make docs responsive

vcarl:
This is a relatively large change to the layout of /docs on mobile. I had to duplicate the left nav, which isn't ideal, but it seemed much easier to manage than trying to get 1 component to fulfill both behaviors.

Screen Shot 2020-04-15 at 6 24 45 PM
Screen Shot 2020-04-15 at 6 25 21 PM


src/components/Documentation/LeftNav.js
```diff
@@ -0,0 +1,84 @@
...
+const LinkEl = styled(Link)`
+  color: ${PALETTE.purpleBlue};
+`;
+const RootLinkEl = styled(Link)`
+  color: ${PALETTE.black80};
+  line-height: 1.75rem;
+  text-decoration: none;
+`;
+
+export const LeftNav = ({
```
vcarl, line 37: I yanked this out of `templates/Documentation` because it was getting large, and I needed to use it in a second location.

src/components/Documentation/MobileLeftNav.js
```diff
@@ -0,0 +1,138 @@
...
+    display: none;
+  }
+`;
+const NavIconSectionEl = styled.div`
+  display: flex;
+  align-items: center;
+`;
+const NavBarEl = styled.div`
+  display: flex;
+  width: 200vw;
```
vcarl, line 33: I feel somewhat clever about this. The menu is actually an overlay that's 2x as wide as the screen. The navbar is 2 navbars next to each other, and when it's open, it moves 1 screen width to the right, exposing the menu.

src/components/Documentation/MobileLeftNav.js
```diff
@@ -0,0 +1,138 @@
...
+  height: 1rem;
+`;
+const MenuButtonEl = styled(BasicButton)`
+  padding: 1rem;
+  margin: 0 -1rem;
+`;
+const OffscreenContainerEl = styled.div`
+  width: 100vw;
+  transform: translate(${(props) => (props.isOpen ? "0" : "-100%")});
+  transition: transform 0.2s;
```
vcarl, line 56: Forgot to update this to use a constant after hardcoding initially, will follow up

src/components/SideNav/SideNav.js
```diff
@@ -29,14 +27,6 @@ const checkIfOverview = (relativePath) => {
   return pathSegments[pathSegments.length - 1] === "index.mdx";
 };
 
-export const SideNav = ({ children, ...props }) => (
```
vcarl, line 32: I realized this is just re-exporting `StickyEl`, so i removed this, updated imports, and renamed StickyEl to `SideNavContainer`

src/components/Documentation/MobileLeftNav.js
vcarl: Since it's an overlay, it's maybe a little too clever about pointer events. They're ignored on the topmost element, then un-ignored on the navbar.

jeesunikim:
Looking great!

src/components/Documentation/SharedStyles.js
```diff
@@ -19,7 +19,7 @@ export const SideNavColumn = styled(Column)`
 export const DividerEl = styled.span`
   display: inline-block;
   width: 0.0625rem;
-  height: 1.5rem;
+  height: 2.75rem;
```
jeesunikim, line 22: This `DividerEl` is also being used in `<CodeExample/>` which has a shorter height (previous 1.5rem) value.

vcarl: Ah good callout, ty

PR #116, opened 2020-04-21
Polish scroll position tracking

vcarl:
Especially on API Reference, this was pretty sub-optimal before. It frequently display a section as being "active" when it was no longer on screen, or when there was a better candidate. I fixed the logic there so it behaves how I originally intended:

Screen Shot 2020-04-17 at 1 28 54 PM

Content becomes "active" when it crosses a boundary near the top/bottom of the page. Some comparisons of when sections begin to be "active":

Screen Shot 2020-04-20 at 4 11 39 PM
Screen Shot 2020-04-20 at 4 12 22 PM

And when scrolling up:

Screen Shot 2020-04-20 at 4 13 13 PM

After fixing that, I noticed that `ScrollRouter` and `SideNav` use similar but not identical methods to track which section is active, meaning that they got out of sync. `ScrollRouter` was tracking the sections themselves, and the sidebar tracking headings. This change overrides the side nav position tracking so that it's all handled by ScrollRouter on API Reference, keeping them in sync.


gatsby-browser.js
```diff
@@ -65,9 +65,3 @@ export const shouldUpdateScroll = ({ routerProps }) => {
   }
   return true;
 };
-
-export const onClientEntry = async () => {
-  if (typeof IntersectionObserver === `undefined`) {
-    await import(`intersection-observer`);
```
vcarl, line 71: We no longer have anything using intersection observer! I tried to use it for this, and got it working, but it didn't improve tracking. Intersection observer reports how much of the element is visible, not how much of the screen it's taking up, so variations in content size were influencing when sections would trigger.

src/components/ApiRefRouting/Route.js
```diff
@@ -13,12 +13,10 @@ export const Route = ({ children, originalFilePath, path }) => {
...
   );
   const ref = React.useRef();
-  const childRef = React.useRef();
   React.useEffect(() => {
-    childRef.current = ref.current.children[0];
-    trackElement(childRef, path);
+    trackElement(ref, path);
     return () => {
-      stopTrackingElement(childRef);
+      stopTrackingElement(ref);
```
vcarl, line 19: By tracking the first child, It ended up tracking the heading. Since the logic I wanted to use is based on watching the top and bottom of the content, that doesn't work as desired. Track the wrapper div.

src/components/ApiRefRouting/ScrollRouter.js
```diff
@@ -18,12 +19,21 @@ let lastScrollPosition = 0;
...
-  const activeNodeRef = React.useRef();
+  const [activeNode, setActiveNode] = React.useState({
+    ref: null,
+    id: initialActive,
+  });
   const trackedElementsRef = React.useRef([]);
   const isScrollingDown = React.useRef(false);
 
+  React.useEffect(() => {
+    contentDom = document.querySelector(`#${DOM_TARGETS.contentColumn}`);
```
vcarl, line 34: This might be a premature optimization, but since this _can_ be selected once, I figured why not 🤷 Especially since my other changes here mean that the effect will be running more frequently.

src/components/ApiRefRouting/ScrollRouter.js
```diff
@@ -92,10 +102,30 @@ export const ScrollRouter = ({ children }) => {
...
+        id: routeMap.get(activeNode),
+      },
+    }),
+    [activeNode],
+  );
 
-  return <Context.Provider value={contextValue}>{children}</Context.Provider>;
+  return (
+    <Context.Provider value={contextValue}>
+      <SideNavProgressContext.Provider value={sideNavContextValue}>
```
vcarl, line 121: This is a little hacky, but short of a full refactor of sidenav + scrollrouter to get them to share some of the underlying code, I couldn't come up with something better. These have very similar behaviors, they just track different elements and respond slightly differently.

src/components/SideNav/Provider.js
```diff
@@ -1,58 +1,54 @@
...
+  const bY = b.ref.current.getBoundingClientRect().top;
   return aY - bY;
 };
 let lastScrollPosition = 0;
 
 export const SideNavProgressContext = React.createContext({});
 
 export const Provider = ({ children }) => {
-  const [activeNode, setActiveNode] = React.useState();
+  const [activeContent, setActiveNode] = React.useState({ id: "", ref: null });
```
vcarl, line 19: The rest of these changes are precipitated by changing `ScrollRouter` to own the SideNav provider as well. It's no longer sufficient to peel an ID off the inner child, so tracking needed to be updated to house a dom node + id

src/components/SideNav/TrackedContent.js
```diff
@@ -6,35 +6,23 @@ import { SideNavProgressContext } from "components/SideNav/Provider";
 
 const El = styled.div``;
 
-export const TrackedContent = ({ children }) => {
+export const TrackedContent = ({ children, identifier }) => {
```
vcarl, line 9: I called this `identifier` and not `id` in an attempt to clarify that it's not a DOM id. Not sure that's super helpful tho

src/helpers/dom.js
```diff
@@ -42,9 +42,9 @@ export const findActiveNode = (possibleNodes, isScrollingDown) => {
     if (!x.current) {
       return false;
     }
-    const { bottom } = x.current.getBoundingClientRect();
+    const { top, bottom } = x.current.getBoundingClientRect();
     return isScrollingDown
-      ? bottom > topEdge && bottom < window.innerHeight
-      : bottom < bottomEdge && bottom > 0;
+      ? top < bottomEdge && top > 0
+      : bottom > topEdge && bottom < window.innerHeight;
```
vcarl, line 48: Relevant change to the "active section" tracking logic. Basically the rest of this PR is because after changing these lines, the side nav and URL got out of sync sometimes.

src/components/SideNav/Provider.js
```diff
@@ -1,58 +1,54 @@
...
 
   React.useEffect(() => {
     const contentDom = document.querySelector(`#${DOM_TARGETS.contentColumn}`);
 
-    const handler = () => {
-      // If we haven't scrolled at least 20 pixels, just bail.
-      if (Math.abs(contentDom.scrollY - lastScrollPosition) < 20) {
+    const handler = throttle(() => {
+      // If we haven't scrolled at least 100 pixels, just bail.
+      if (Math.abs(contentDom.scrollY - lastScrollPosition) < 100) {
```
piyalbasu, line 33: super small: this 100 "threshold" is used in scrollrouter, too - might be nice to save it as a const somewhere

gatsby-browser.js
piyalbasu: maybe i'm misunderstanding, but we don't need to polyfill the IntersectionObserver instance in `useElementPosition`?

vcarl: That was a mis-commit! I didn't end up using it, got swept up when I merged master.

src/components/SideNav/Provider.js
vcarl: Yeah there's a lot of room to unify the behavior there, I think, but it started to look like a big effort with relatively little benefit to the site. Maybe one day.

PR #122, opened 2020-04-27
Updating README.md for our reader

jeesunikim:


README.md
```diff
@@ -49,15 +118,349 @@ src/
...
+## Custom Components
+
+Custom (React) components that are being used throughout Documentation and API
+Reference.
+
+### AttributeTable
+
+<img src="./readme-imgs/attribute-table.png" alt="AttributeTable Component" width="600"/>
+
+`<AttributeTable/>` is currently used in API Reference. Nested Table is allowed.
```
vcarl, line 230: I'm not sure these "X is currently used on Y" is helpful information for content editors, and it seems likely to get out of date. I don't feel strongly but maybe these lines should be removed, thoughts?

I'm not sure exactly what "Nested Table is allowed" is meant to convey, maybe this should explain how nested attributes turn into "child attributes" that are collapsed by default?

README.md
```diff
@@ -49,15 +118,349 @@ src/
...
+<img src="./readme-imgs/code-example.png" alt="CodeExample Component" width="600"/>
+
+`<CodeExample/>` is currently used in both API Reference and Documentation. It
+is using [Prism syntax highlighting library](https://prismjs.com/).
+
+You can include snippets for more than one language. See an example of including
+a snippet for `curl` and `JavaScript` below.
+
+For example,
+[Seeting Up Test Server](https://developers.stellar.org/docs/enabling-deposit-and-withdrawal/setting-up-test-server/)
```
vcarl:
```suggestion
[Setting Up Test Server](https://developers.stellar.org/docs/enabling-deposit-and-withdrawal/setting-up-test-server/)
```

README.md
```diff
@@ -49,15 +118,349 @@ src/
...
+```
+
+### Alert
+
+<img src="./readme-imgs/alert.png" alt="Alert Component" width="712"/>
+
+`<Alert/>` is currently used in Documentation.
+
+For example,
+[Seeting Up Test Server](https://developers.stellar.org/docs/enabling-deposit-and-withdrawal/setting-up-test-server/)
```
vcarl:
```suggestion
[Setting Up Test Server](https://developers.stellar.org/docs/enabling-deposit-and-withdrawal/setting-up-test-server/)
```

README.md
```diff
@@ -49,15 +118,349 @@ src/
...
+````
+import { CodeExample } from "components/CodeExample";
+
+<CodeExample title="Example Request">
+
+```curl
+curl "https://horizon.stellar.org/ledgers/27147222/transactions?limit=2"
+```
+
+```JavaScript
```
vcarl, line 306: Should this be `js`? I thought the language here would be the key in the CODE_LANGS lookup, not the value

jeesunikim: Either works

README.md
jeesunikim: Good call. I will mention "X is currently used on Y" for `<ExampleResponse/>` that is being used only in API Reference. I will add a better explanation for the latter.

jeesunikim: I think I want to imply where each component is being used so a reader can find it easier if he/she wants to see it in action (but at the same time, I added a link..)

jeesunikim: Removed!

PR #124, opened 2020-05-01
Fix and consolidate link behavior

vcarl:
Links were a little busted in a couple of small ways, and while repairing them I realized there was an opportunity to bring some of the logic together. I also added some helpers with tests so we can be a little more confident that these are working correctly.

### DIFF
--- a/buildHelpers/routes.js
+++ b/buildHelpers/routes.js
@@ -30,11 +30,16 @@ const DOCS_ROOT = "docs";
 
 const buildPathFromFile = (filePath) => {
   const pathRegex = /(.*)\..*$/;
-
   // Strip `index` so that `index.mdx` files come through with just their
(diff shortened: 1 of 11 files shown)


jeesunikim:
Just noticed that some of the links got an incorrect url. For example, Links such as `stellar-core`'s `https://developers-11798b5.previews.kube001.services.stellar-ops.com/docs/glossary/stellar-core/` and `Ledger`'s link (https://developers-11798b5.previews.kube001.services.stellar-ops.com/docs/glossary/ledgers/) should be a single `ledger` not `ledgers`.

Not an issue with this PR. Just something I noticed 👀, @rice2000

vcarl:
I think this is out of date relative to master, as well. There might be some content issues in this preview that have already been fixed.

PR #127, opened 2020-05-07
Bugfix/typeform - Fixes for bugs reported via Typeform

jeesunikim:
Fixes for bugs reported via Typeform

* [Subscribe Button Overlaps](https://app.asana.com/0/1119309091112078/1173687524197297)
* [Triple Scrollbar on Mac and Linux](https://app.asana.com/0/1119309091112078/1173720962232793)

### DIFF
--- a/src/basics/GlobalStyles.js
+++ b/src/basics/GlobalStyles.js
@@ -37,10 +37,6 @@ export const GlobalStyles = createGlobalStyle`
   twitter-widget {
     margin: auto;
   }
-  ::-webkit-scrollbar {
-    width: 0px;  /* Remove scrollbar space */
-    background: transparent;  /* Optional: just make scrollbar invisible */
-  }
 
 ${FONTS.map(
   (font) => css`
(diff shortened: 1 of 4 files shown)


src/basics/GlobalStyles.js
```diff
@@ -37,6 +37,11 @@ export const GlobalStyles = createGlobalStyle`
   twitter-widget {
     margin: auto;
   }
+  ::-webkit-scrollbar {
```
vcarl, line 40: Verrrry nervous about globally hiding scrollbars. I notice that the entire page scrolls a little bit, I think there's some extra height coming in from the sidebar

vcarl: I looked at this earlier but hadn't committed a fix, I think `SideNavContainer` has the wrong height. It's doing `calc(100vh - 3.75rem)` but the logo section is 72px = 4.5rem.

jeesunikim: How about adding this to the JSON responses component and I will find the extra height?

vcarl: This could def make sense on those docs elements

jeesunikim: Updated!

PR #135, opened 2020-05-11
alert/remove-space > removing an extra space for markdown to render correctly

jeesunikim:


vcarl:
I'm not sure what tools are available, but it would be cool if we could somehow lint for the known gotchas we have, or dig into why exactly we need this extra line for it to be parsed by the markdown loader and open a PR to change it.

PR #141, opened 2020-05-14
Improve API Reference loading performance

vcarl:
It turns out that Prism is generates a ton of MDX nodes, and because of how Gatsby loads page data, that was inflating `page-data.json` by a ton. About somewhere between 50-70% of the page bulk on /api was directly attributable to syntax highlighting. It looks like I was able to shift that over to render clientside without it being too disruptive to the end user, which cut the amount of data we're loading by a lot

The other major win is adding a script to the end of `<body>` that immediately scrolls the current section into view. It's not as flawlessly perfect as Stripe's is (I still don't know how they do that!) but it's pretty dang good. The page still chugs a little, but that's what happens when you load > 2MB of html.


src/components/CodeExample.js
```diff
@@ -295,9 +300,25 @@ export const CodeExample = React.forwardRef(function CodeExample(
   { title, children, href, ...props },
   ref,
 ) {
+  const codeSnippets = React.useMemo(
```
jeesunikim, line 303: Nice

jeesunikim:
Great work! 

Now that `availableLangs` is getting returned values from new `codeSnippets` which returns an array of code language values such as `["curl", "JavaScript"]` vs. previous one `["curl", "JavaScript"]`.

Can we update the following `{CODE_LANGS[langValue]}` to `{langValue}` - line 241.

`{availableLangs.map((langValue) => (<option key={langValue} value={langValue}>{CODE_LANGS[langValue]}</option>))}`

src/basics/Prism.js
```diff
@@ -1,40 +1,91 @@
...
+  string: DARK_PALETTE.green,
+  function: DARK_PALETTE.yellow,
+  boolean: DARK_PALETTE.yellow,
+  object: DARK_PALETTE.yellow,
+  comment: DARK_PALETTE.grey,
+  variable: DARK_PALETTE.light,
+  htmlTag: DARK_PALETTE.grey,
+  cssVariable: DARK_PALETTE.yellow,
+  cssSelector: DARK_PALETTE.green,
+};
```
vcarl, line 62: Worth highlighting. I revamped how we're applying styles so that we can more easily adjust them in the future. New colors from Charles didn't look good on the code examples, so it made sense to split them apart. We might need to double check these as we add content with new languages.

zech564 (commented):
A lot to look at

PR #143, opened 2020-05-20
outline-tracking-docs > polish tracking on /docs; separate logic for click and scroll

jeesunikim:
[Docs outline active section tracking shows previous section](https://app.asana.com/0/1130719348309291/1175224232695578)
- Unlike /api, /docs was only tracking the headings instead of the sections themselves. Consolidated the flat structure to group sections to track better. This solved a lot of the issues we had with /docs' highlighting bugs
- Added a new state to detect if a user clicks a nav link, highlight that link regardless the position of the content by overwriting it.
- Previously, when a user scrolled up we highlighted the last element in view. After adding a `click and highlight` state, it had an unintended result. See below gif.

With highlighting the last element when scrolling up - after a user clicks `Run an API Server` and scrolls up, it highlights `Software and SDKs` instead of `Run a Core Node`.

Before:
before-fixing

After updating it to highlighting the first shown element instead of the last when scrolling up:
after-fixing

This matches the behavior on our stellar.org's side nav [highlighting](https://www.stellar.org/start).

This doesn't seem to affect /api. But I will double check again.

### DIFF
--- a/src/components/SideNav/Provider.js
+++ b/src/components/SideNav/Provider.js
@@ -17,6 +17,8 @@ export const Provider = ({ children }) => {
   const [activeContent, setActiveNode] = React.useState({ id: "", ref: null });
   const [trackedElements, setTrackedElements] = React.useState([]);
 
+  const [isNavClicked, onNavClick] = React.useState(false);
+
   // We need to search just the refs pretty frequently, so memoize it so we
   // don't generate a ton of garbage.
   const elementRefs = React.useMemo(() => trackedElements.map((e) => e.ref), [
@@ -29,11 +31,18 @@ export const Provider = ({ children }) => {
       if (Math.abs(window.scrollY - lastScrollPosition) < 100) {
         return;
       }
+
       const isScrollingDown = window.scrollY > lastScrollPosition;
       lastScrollPosition = window.scrollY;
 
       const newActiveRef = findActiveNode(elementRefs, isScrollingDown);
+
+      if (isNavClicked) {
+        return;
+      }
+
       const newActiveNode = trackedElements.find((e) => e.ref === newActiveRef);
+
       if (newActiveNode && newActiveNode !== activeContent) {
         setActiveNode(newActiveNode);
       }
@@ -41,10 +50,11 @@ export const Provider = ({ children }) => {
 
     window.addEventListener("scroll", handler);
     handler();
+
     return () => {
       window.removeEventListener("scroll", handler);
     };
-  }, [activeContent, elementRefs, trackedElements]);
+  }, [activeContent, elementRefs, trackedElements, isNavClicked]);
 
   const trackElement = React.useCallback(
     (ref) => {
@@ -64,8 +74,16 @@ export const Provider = ({ children }) => {
       activeContent,
       stopTrackingElement,
       trackElement,
+      setActiveNode,
+      onNavClick,
     }),
-    [activeContent, stopTrackingElement, trackElement],
+    [
+      activeContent,
+      stopTrackingElement,
+      trackElement,
+      setActiveNode,
+      onNavClick,
+    ],
   );
 
   return (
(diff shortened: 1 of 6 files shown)


src/helpers/documentation.js
```diff
@@ -376,8 +376,38 @@ export const groupByCategory = (referenceDocs) => {
   }, {});
 };
 
+/**
+ * consolidateToSection is a reducer function to consolidate
+ * all of subsequent items after <h2/> into an array of section
+ * For <TrackedContent/> to surround its H2 and the content for /docs
+ */
+export const consolidateToSection = () => {
```
vcarl, line 384: Ah so this is looking for the H2s, then marking all content between them as a "section" of content? Nice

src/components/SideNav/Provider.js
```diff
@@ -29,22 +31,30 @@ export const Provider = ({ children }) => {
       if (Math.abs(window.scrollY - lastScrollPosition) < 100) {
         return;
       }
+
       const isScrollingDown = window.scrollY > lastScrollPosition;
       lastScrollPosition = window.scrollY;
 
       const newActiveRef = findActiveNode(elementRefs, isScrollingDown);
+
+      if (isNavClicked) {
```
jeesunikim, line 40: Overwriting our track and scroll behavior here.

src/helpers/dom.js
```diff
@@ -7,18 +6,18 @@ import { ScrollToPlugin } from "gsap/ScrollToPlugin";
...
-  const { offset = 0, duration = 0.55 } = options;
+export const smoothScrollTo = (node, onCompleteFn) => {
+  const offset = 0;
+  const duration = 0.55;
 
   TweenLite.to(window, duration, {
     scrollTo: {
       y: node.offsetTop + offset,
     },
+    onComplete: onCompleteFn,
```
jeesunikim, line 17: onCompleteFn is where we set `isNavClicked` to false once the smooth scroll behavior is completed.

src/helpers/dom.js
```diff
@@ -28,20 +27,27 @@ const reverse = memoize((arr) => arr.slice().reverse());
  */
 export const findActiveNode = (possibleNodes, isScrollingDown) => {
   const topEdge = window.innerHeight * 0.125;
-  const bottomEdge = window.innerHeight * 0.875;
+  const bottomEdge = window.innerHeight * 0.5;
```
jeesunikim, line 30: I reduced the `bottomEdge` value to match the footer we have

src/templates/Documentation.js
```diff
@@ -106,41 +110,66 @@ const ModifiedEl = styled.div`
...
+
+  return (
+    <NavItemEl
+      isActive={isActive}
+      onClick={(e) => {
+        e.preventDefault();
+        onNavClick(true);
+        setActiveNode(document.getElementById(id));
+        smoothScrollTo(document.getElementById(id), () => {
+          onNavClick(false);
```
jeesunikim, line 128: Where I'm toggling `onNavClick` value so we can stop scroll tracking behavior

vcarl: I think it'd be nice to keep this logic contained within the sidenav provider—since this is a page template, the the bookkeeping details of when to toggle state in the sidenav seem too granular. Not sure the tech details work out but something like `focusNavItem(id)` here would be awesome

jeesunikim: Not sure about putting this in the sidenav provider since both `/api` and `/docs` use different `renderItems` for its `SideNavBody`. `/api` uses `ScrollRouter` in order to change its window state when the nav is clicked where `/docs` doesn't do that. Also `smoothScrollTo` is only used in `/docs`.

vcarl: Hm maybe it's not super feasible then. This is a relatively minor note, not worth blocking the PR over

PR #144, opened 2020-05-21
Polish tooltips

vcarl:
Took a few false starts, but this is a little more durable and prettier.

### DIFF
--- a/src/components/CodeExample.js
+++ b/src/components/CodeExample.js
@@ -3,8 +3,9 @@ import styled from "styled-components";
 import PropTypes from "prop-types";
 import { CopyToClipboard } from "react-copy-to-clipboard";
 
-import ExternalLinkIcon from "assets/icons/icon-external-link.svg";
-import { CopyIcon } from "basics/Icons";
+import { BasicButton } from "basics/Buttons";
+import { CopyIcon, ExternalLinkIcon } from "basics/Icons";
+import { Code } from "basics/Text";
 
 import { PALETTE, FONT_FAMILY, FONT_WEIGHT } from "constants/styles";
 
@@ -16,7 +17,6 @@ import { Link } from "basics/Links";
 import { Select } from "basics/Inputs";
 
 import { Tooltip } from "components/Tooltip";
-import { Code } from "basics/Text";
 
 const CODE_LANGS = {
   bash: "bash",
--- a/src/components/CodeExample.js
+++ b/src/components/CodeExample.js
@@ -42,7 +42,6 @@ const LangSelect = styled(Select)`
 
   select {
     display: inline-block;
-    margin: 0 1rem;
     padding: 0.5rem;
     font-size: 0.75rem;
     border: none;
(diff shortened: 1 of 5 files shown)


src/components/Tooltip.js
```diff
@@ -1,97 +1,116 @@
...
+    if (!isOpen || !tooltipRef.current) {
+      return;
+    }
+    const { right, top } = wrapperRef.current.getBoundingClientRect();
+    const newRight = window.innerWidth - right;
+    // We need height + some because of the extra space for the little pointer
+    // bit. That's .25rem away
+    const newTop = window.pageYOffset + top - (size.height + 4);
+    tooltipRef.current.style.right = `${newRight}px`;
+    tooltipRef.current.style.top = `${newTop}px`;
```
vcarl, line 81: I switched this to imperatively set the size instead of setting state to avoid another render pass.

The major change here is that I switched it from `left:` to `right:`, which lets us more easily have dynamically sized content without the position changing

src/components/CodeExample.js
```diff
@@ -248,38 +241,27 @@ const CodeSnippet = ({ codeSnippets, title, href }) => {
...
+              onCopy={() => !isCopied && setCopy(true)}
             >
-              <CopyIcon />
-            </CopyIconWrapper>
-          </CopyToClipboard>
+              <CopyWrapperEl>
+                <CopyIcon />
+              </CopyWrapperEl>
+            </CopyToClipboard>
+          </Tooltip>
```
vcarl, line 253: By wrapping the hover target, I could move the position detection logic inside `Tooltip`. The logic is basically the same, but now it's contained within `Tooltip`, which is nice.

src/components/CodeExample.js
```diff
@@ -111,25 +110,32 @@ const CodeExampleEl = styled.div`
   }
 `;
 
-const CopyIconWrapper = styled.div`
-  display: block;
-  position: relative;
+const CopyWrapperEl = styled.div`
   cursor: pointer;
+  padding: 0.75rem 1rem;
```
vcarl, line 115: I thought the click-to-copy target was a little small, so I moved the padding off `TitleEl` and onto the copy button. Had to shuffle some other margins/padding to make it work

src/components/Tooltip.js
```diff
@@ -1,97 +1,111 @@
...
 
 import { PORTAL_TARGETS } from "constants/domNodes";
 import { PALETTE, Z_INDEXES } from "constants/styles";
 
 import { renderToPortal } from "helpers/renderToPortal";
 
+// For the arrow, it should be 4px (.25rem) away from the edge. Since it's a box
+// rotated 45 degrees, calculate the length of 1 side of the box.
+// a^2 + b^2 = c^2
+const hypotenuse = (0.25 ** 2 + 0.25 ** 2) ** 0.5;
```
piyalbasu, line 13: Mrs. Caulfield, my high school geometry teacher, salutes you

jeesunikim:
tooltip

I noticed all of tooltips are being generated instead of just one that is being used.

src/components/Tooltip.js
vcarl: I DEFINITELY did not try to use SOHCAHTOA first

vcarl:
Yeah, but I think that's okay 🤷 Having them all in the markup simplifies logic a little. I was running into a pop-in issue where it would render in the wrong place, get measured, then move to the right location.

PR #145, opened 2020-05-22
Make expansion timing more consistent

vcarl:
With how we we doing animated expansions before, the timing varied because the exact height wasn't known in advance. By providing information about the height to the component in question, it will always run for the length of time specified.


jeesunikim:
I'm seeing a glitch when expanding side navigation on /docs locally. I attached a video in asana [ticket](https://app.asana.com/0/1130719348309291/1175224232695565). Does this happen to you too, @piyalbasu? I wonder if it's just me.

vcarl:
I'm seeing it too, definitely something up. We pretty consistently have some weird issues with font loading, I'm not sure why. I assume it has something to do with `GlobalStyles` somehow refreshing, but afaik styled-components/gatsby has been set up correctly 🤔

jeesunikim:
I wonder why it's more noticeable now than it was before

vcarl:
I noticed this doesn't account for when the expanded section changes height after mounting, like when an API reference sub-section enters the view. Gonna have to rework it
Screen Shot 2020-05-26 at 11 47 29 AM
Screen Shot 2020-05-26 at 11 47 21 AM

PR #147, opened 2020-05-27
Added the latest prettier in order to prettify mdx files

jeesunikim:
* Removed `*.mdx` from `.prettierignore`
* Installed the latest `prettier` for [MDX support](https://github.com/prettier/prettier/issues/7041)
* It works pretty well. It removes an empty space in a blanket line (that is required in order to render JSX component) if a user adds it before the closing mark. The content writer still needs to make sure that there's an empty line in between component declaration.
* Prettified all the files **EXCEPT** the following files:
`content/api/resources/accounts/object.mdx`
`content/api/resources/effects/types.mdx`

Those files' `<AttributeTable/>` got an empty list which prettier prettify and breaks the code. I'm going to investigate whether we can use something like `null` instead of an empty list

example

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^24.9.0",
     "lodash": "^4.17.15",
-    "prettier": "^1.17.0",
     "pretty-quick": "^1.10.0",
     "prismjs": "^1.17.1",
     "prop-types": "^15.7.2",
@@ -99,5 +98,8 @@
   },
   "homepage": "https://stellar.org",
   "license": "MIT",
-  "prettier": "@stellar/prettier-config"
+  "prettier": "@stellar/prettier-config",
+  "devDependencies": {
+    "prettier": "prettier/prettier"
+  }
 }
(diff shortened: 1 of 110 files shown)


package.json
```diff
@@ -99,5 +98,8 @@
   },
   "homepage": "https://stellar.org",
   "license": "MIT",
-  "prettier": "@stellar/prettier-config"
+  "prettier": "@stellar/prettier-config",
+  "devDependencies": {
+    "prettier": "prettier/prettier"
```
vcarl, line 103: I've started avoiding `devDependencies` on apps, because we have to install deps + devDeps anyway for a production build, so it doesn't seem like a useful distinction. devDeps useful for libraries so they don't put more on disk than what's needed

vcarl: Oh also, looks like this is pulling from github? Do we need to avoid the npm version?

jeesunikim: It looks like our `.prettierrc.js` is being ignored. Its `proseWrap` is set to "never"`, but still wraps it. Investigating it..

jeesunikim: @vcarl, I figured it out. Making a new PR and will add details. Closing this.

PR #148, opened 2020-05-29
Update prettier for .mdx

jeesunikim:
We had two [prettier configuration](https://prettier.io/docs/en/configuration.html)

1. A "prettier" key in your `package.json` file.
2. `.prettierrc.js`

Even though we had `proseWrap: "never"` in `.prettierrc.js`, it was getting `proseWrap: "always"` from our [stellar prettier config](https://github.com/stellar/product-conventions/blob/master/packages/prettier-config/prettier.config.js) set in our `package.json`.

I also updated our `.prettierrc.js` to use prettier's [Configuration Overrides](https://prettier.io/docs/en/configuration.html#configuration-overrides) - I assume we want `proseWrap` for the files that aren't `.mdx`

This is nice because prettify prettifies our code snippets within `<CodeExample/>` and delete unnecessary spacings and converted `*` to `-` to use the correct markdown for `list`.

[Major Changes that aren't formatted files](https://github.com/stellar/new-docs/pull/148/commits/aa261367398a2679d786666294bef98800387433)
`.prettierignore`
`.prettierrc.js`
`package.json`


.prettierrc.js
```diff
@@ -2,5 +2,12 @@ module.exports = {
...
   // diffs. Override the default for a better experience.
-  proseWrap: "never",
+  overrides: [
+    {
+      files: "*.mdx",
+      options: {
+        proseWrap: "never",
+      },
+    },
+  ],
```
vcarl, line 12: Dang I thought spreading the stellar config would have been fine. Thanks for digging in

PR #151, opened 2020-06-05
Replaced 'window.scrollTop' with 'window.scrollY' and add an onClick even to highlight navigation

jeesunikim:
[Ticket: API Ref active section sometimes doesn't line up after clicking a link](https://app.asana.com/0/1130719348309291/1175224232695574)

On `/api`, it was using `window.scrollTop` instead of `window.scrollY` to detect its scrolling measurement. `window.scrollTop` does not exist so it was returning `undefined`. Fixing this fixed the scroll and highlighting issue.

Unlike `/docs`, when a user clicks a link on side navigation on `/api`, its behavior changes window's history state and instead of showing smooth scroll up and down, it gets a user to the section right away - which means `isScrollingDown.current` was returning `undefined` and using the logic of scrolling up instead of scrolling down from `findActiveNode` from `/helpers/dom`. 

When clicking the link, its content's top is set to `-4`, I adjusted the number value for `findActiveNode` and set it to use `isScrollingDown` logic.

`content/api/resources/ledgers/index.mdx `
- Corrected the link to the /docs for ledgers. Will go through all of the links to make sure they are linked correctly

### DIFF
--- a/src/components/ApiRefRouting/ScrollRouter.js
+++ b/src/components/ApiRefRouting/ScrollRouter.js
@@ -48,7 +48,7 @@ export const ScrollRouter = ({ children, initialActive = "" }) => {
         isNavClicked ? true : isScrollingDown.current,
       );
 
-      if (isNavClicked) onNavClick(false);
+      if (isNavClicked) onNavClick(!isNavClicked);
 
       // If we've found an active node and it's not the same one as we had
       // before, update the route.
--- a/src/components/ApiRefRouting/ScrollRouter.js
+++ b/src/components/ApiRefRouting/ScrollRouter.js
@@ -23,6 +23,7 @@ export const ScrollRouter = ({ children, initialActive = "" }) => {
     ref: null,
     id: initialActive,
   });
+  const [isNavClicked, onNavClick] = React.useState(false);
   const trackedElementsRef = React.useRef([]);
   const isScrollingDown = React.useRef(false);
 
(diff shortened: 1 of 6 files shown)


src/helpers/dom.js
```diff
@@ -45,7 +45,7 @@ export const findActiveNode = (possibleNodes, isScrollingDown) => {
       // Display the last item when scrolled all the way to the bottom
       return isReachedBottom
         ? bottom + bottomEdge > window.innerHeight - topEdge
-        : top < bottomEdge && top > 0;
+        : top < bottomEdge && top >= -5;
```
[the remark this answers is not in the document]
vcarl: Hmm why is it -4? It would be nice to keep this at 0, since it's supposed to be "top of the screen"

src/templates/ApiReference.js
```diff
@@ -187,7 +187,14 @@ const NavItem = ({ isActive, forwardedRef, children, depth }) => {
   }, [isActive, parentDom, isScrollingDown, isMobile]);
 
   return (
-    <NavItemEl isActive={isActive} depth={depth} ref={itemRef}>
+    <NavItemEl
+      isActive={isActive}
+      depth={depth}
+      ref={itemRef}
+      onClick={() => {
+        onNavClick(true);
```
vcarl, line 195: This is the only use of `onNavClick` right? The `true` doesn't really have a meaning here, it's an implementation detail of `ScrollRouter`, I think it would be clearer if the function coming out of context were called as `onNavClick()`

jeesunikim: `onNavClick` is also used in `ScrollRouter` to set it as `false` for `findActiveNode` to use the `isScrollingDown` approach.

piyalbasu: what if `onNavClick` is called `setIsNavClicked` instead? it'd clear that it's setting a boolean `isNavClicked` in `ScrollRouter`, so passing `true` to it here would make sense

src/components/ApiRefRouting/ScrollRouter.js
```diff
@@ -36,16 +37,19 @@ export const ScrollRouter = ({ children, initialActive = "" }) => {
...
+      isScrollingDown.current = window.scrollY > lastScrollPosition;
+      lastScrollPosition = window.scrollY;
 
       const newActiveNode = findActiveNode(
         trackedElementsRef.current,
-        isScrollingDown.current,
+        isNavClicked ? true : isScrollingDown.current,
       );
+
+      if (isNavClicked) onNavClick(!isNavClicked);
```
piyalbasu, line 51: looks like this can just be `onNavClick(false)` since that's the end result you're looking for

src/templates/ApiReference.js
piyalbasu: oh, i see that it's following the convention of calling it `onNavClick` from `SideNavProgressContext`. 

I think calling it `setIsNavClicked` makes it more clear that this function requires a boolean, but if you don't want to change this convention everywhere, i understand

jeesunikim: Updated the names! Will merge.

PR #155, opened 2020-06-15
Rice2000 add start

rice2000:
- add content to where to start
- add content to software and sdks


content/docs/start/introduction.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: Introduction
+order: 0
+---
+
+import { CodeExample } from "components/CodeExample";
```
jeesunikim, line 6: It looks like this isn't being used in this .mdx?

PR #158, opened 2020-06-16
Fix hash links, API ref link speed, docs expando

vcarl:


src/templates/ApiReference.js
```diff
@@ -120,7 +120,7 @@ const NavItemEl = styled.div`
 
 const StyledLink = components.a;
 
-const NavLinkEl = styled(BasicLink)`
+const NavLinkEl = styled(Link)`
```
vcarl, line 123: Turns out this is why links on api ref were slow! Was doing a regular gastby route transition, totally ignoring the optimizations we'd out in place

src/components/Documentation/Articles.js
```diff
@@ -177,7 +177,7 @@ const Articles = ({
         </TopicExpander>
       )}
 
-      <ExpandedSection isExpanded={isExpanded}>
+      <ExpandedSection isOpened={isExpanded}>
```
vcarl, line 180: Missed this in my last PR. Left nav in regular docs weren't opening, oops

src/templates/ApiReference.js
jeesunikim: Nice!!

PR #161, opened 2020-06-17
Welcome page > to have Read More link instead of having H2 to be linked

jeesunikim:
Our [docs](https://developers.stellar.org/docs/)' `Welcome Page` is the only page with H2 that has a link. I spoke to Charles about its styling and Charles suggested to have a **Read More** CTA button at the bottom instead of H2 with a link

**Preview:**
Old way
old-h2-links

New way
new-readmore

### DIFF
--- a/content/docs/index.mdx
+++ b/content/docs/index.mdx
@@ -3,7 +3,7 @@ title: Welcome
 order: -100
 ---
 
-import { CodeExample } from "components/CodeExample";
+import { ReadMore } from "components/ReadMore";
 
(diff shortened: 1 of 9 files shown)


content/docs/index.mdx
```diff
@@ -15,34 +15,50 @@ In addition to the docs here, which are guides to building on Stellar, you'll al
 
 A rundown of what's here:
 
-## [Tutorials](./tutorials/create-account.mdx)
+## Tutorials
 
 If you’re new to Stellar and want to get an overview of the network, this is where to start. The early tutorials will show you how to do some basic things like create an account and make payments. The later tutorials cover more advanced topics like Stellar smart contracts. If you want to issue an asset or build an app, you're better served checking the sections dedicated to those paths.
 
-## [Issue Assets](./issuing-assets/index.mdx)
+<ReadMore url="./tutorials/create-account" />
```
vcarl, line 22: This seems fine, but it's a mild bummer that this isn't a "real" link that will work within github. Since we're not doing this frequently it seems okay though

PR #163, opened 2020-06-18
Update control-asset-access.mdx

rice2000:
- remove inaccurate sentence
- fix headers so they're more useful


jeesunikim:
👀Up early!

PR #168, opened 2020-06-22
Add metrics and performance tracking

vcarl:
Since we have some concerns about performance, it's important we have performance metrics being emitted somewhere. And I'm not sure what's up with the old Google analytics snippet, but it didn't appear to be loaded. New snippet seems to work

### DIFF
--- a/src/components/layout/LayoutBase.js
+++ b/src/components/layout/LayoutBase.js
@@ -4,6 +4,10 @@ import styled from "styled-components";
 import Helmet from "react-helmet";
 
 import { FONTS } from "constants/fonts";
+import { PERFORMANCE_MARKS } from "constants/performanceMarks";
+
+import { emitMetric } from "helpers/metrics";
+import { mark, measure } from "helpers/performance";
 
 import { BasicLink } from "basics/Links";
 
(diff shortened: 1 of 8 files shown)


src/components/layout/LayoutBase.js
```diff
@@ -37,31 +41,41 @@ export const LayoutBase = ({
...
-);
+}) => {
+  mark(PERFORMANCE_MARKS.layout);
+  React.useEffect(() => {
+    const timing = measure(PERFORMANCE_MARKS.layout);
+    emitMetric("page rendered", {
+      ...timing,
+      page: window.location.pathname,
+    });
+  });
```
vcarl, line 52: Only use of metrics at the moment. Figure this is a good, simple data point that gets to interactivity and overall time to load, where it's within our control. There's a ton of HTML, so we can't do too much about overall page size

vcarl: `timing` here is `{ duration, hasHighPrecision }`

src/helpers/metrics.js
```diff
@@ -0,0 +1,37 @@
+import throttle from "lodash/throttle";
+import { AMPLITUDE_KEY } from "constants/env";
+
+const METRICS_ENDPOINT = "https://api.amplitude.com/2/httpapi";
+let cache = [];
+
+const uploadMetrics = throttle(() => {
```
vcarl, line 7: [Amplitude's SDK is like 60kb](https://bundlephobia.com/result?p=amplitude-js@6.2.0)! We just need some dead simple metrics with no user tracking, so I threw together a shitty little uploader to hit the HTTP API directly.

src/helpers/performance.js
```diff
@@ -0,0 +1,44 @@
...
+  performance.clearMarks(perfMark);
+  performance.clearMeasures(perfMark);
+};
+let internalMark = performance?.mark.bind(performance);
+let internalMeasure = (perfMark) => {
+  const { duration } = performance.measure(null, perfMark);
+  clear(perfMark);
+  return { time: duration, hasHighPrecision };
+};
+if (!hasHighPrecision) {
```
vcarl, line 17: This is some annoyingly obscure looking code, but unfortunately the performance API isn't consistently available. This exposes a `mark` and `measure` function to do simple timers (the perf api offers a lot more complex stuff, but start/stop seems sufficient for most cases), so this checks for if it's available and provides a shitty but functional timer stub if not.

src/html.js
```diff
@@ -1,13 +1,17 @@
...
+import { IS_BUILD } from "constants/env";
 
 export default function HTML(props) {
   return (
     <html {...props.htmlAttributes}>
       <head>
         <meta charSet="utf-8" />
         <meta httpEquiv="x-ua-compatible" content="ie=edge" />
+        {IS_BUILD && (
+          <script async src="https://www.google-analytics.com/analytics.js" />
```
vcarl, line 13: No idea why the old analytics snippet wasn't loading, because all it does is append this script to head. Since that's all it's doing anyway (the weird `i,s,o,g,r,a,m` snippet), I just added it directly.

src/helpers/performance.js
```diff
@@ -0,0 +1,44 @@
...
+  const marks = {};
+  internalMark = (perfMark) => {
+    marks[perfMark] = Date.now();
+  };
+  clear = (perfMark) => {
+    marks[perfMark] = undefined;
```
piyalbasu, line 23: this may not be a possible use case, but should this be set to `0` on `clear` since it might possibly be subtracted from `Date.now()` in `internalMeasure`

vcarl: Hm fair point, but that would only break if `mark` isn't called first, which would break before `clear` is called as well. I'm okay leaving this as is for now but if we end up stealing this in future projects it's something to consider.

PR #169, opened 2020-06-22
using reach router's navigate for hash links

jeesunikim:
**Problem:**
hash-links-bug-480

On /docs' [Run Core Node - configuring](https://developers.stellar.org/docs/run-core-node/configuring/), for example, if you click links that are targeting hash links. They won't work most of the time (some of the hash links I clicked for the first time worked; I've tried multiple times on incognito, but happens randomly).

Gatsby's [Recommendations for programmatic, in-app navigation](https://www.gatsbyjs.org/docs/gatsby-link/#recommendations-for-programmatic-in-app-navigation) recommends to use an `anchor tag` (which what we were already using) or `reach/router`'s `navigate`.

I implemented `navigate` for hash links and this fixes the problem. But it does feel hacky since anchor tag should work so I'd like you guys' opinions.

It looks like there are scroll handling issues within Gatsby
[Several Fixes for Scroll Handling and Restoration](https://github.com/gatsbyjs/gatsby/pull/24306)
[Fix gatsby-react-router-scroll](https://github.com/gatsbyjs/gatsby/pull/25161)


content/docs/run-core-node/configuring.mdx
```diff
@@ -19,7 +19,7 @@ This doc works best in conjunction with concrete config examples, so as you read
...
 
-- If you want to connet to the public network, check out this [public network config for a Full Validator](https://github.com/stellar/packages/blob/master/docs/examples/pubnet-validator-full/stellar-core.cfg). It includes a properly crafted quorum set with all the current [Tier 1 validators](./tier-1-orgs.mdx), which is a good place to start for most configurations. This node is set up to both [validate](#Validating) and write history to a [public archive](./publishing-history-archives.mdx), but you can disable either feature to adjust this config so it's a little lighter.
+- If you want to connet to the public network, check out this [public network config for a Full Validator](https://github.com/stellar/packages/blob/master/docs/examples/pubnet-validator-full/stellar-core.cfg). It includes a properly crafted quorum set with all the current [Tier 1 validators](./tier-1-orgs.mdx), which is a good place to start for most configurations. This node is set up to both [validate](#validating) and write history to a [public archive](./publishing-history-archives.mdx), but you can disable either feature to adjust this config so it's a little lighter.
```
vcarl, line 22: GitHub is not helping me spot the change here, but I do see a typo so I guess it works out. `connet -> connect`
```suggestion
- If you want to connect to the public network, check out this [public network config for a Full Validator](https://github.com/stellar/packages/blob/master/docs/examples/pubnet-validator-full/stellar-core.cfg). It includes a properly crafted quorum set with all the current [Tier 1 validators](./tier-1-orgs.mdx), which is a good place to start for most configurations. This node is set up to both [validate](#validating) and write history to a [public archive](./publishing-history-archives.mdx), but you can disable either feature to adjust this config so it's a little lighter.
```

src/basics/Links.js
```diff
@@ -155,7 +155,7 @@ export const Link = ({ href, newTab, ...props }) => {
       return <BasicLink href={url} {...props} />;
     case LINK_DESTINATIONS.docs:
     case LINK_DESTINATIONS.hash:
-      return <BasicLink href={url} {...props} />;
+      return <BasicLink onClick={() => navigate(url)} href={url} {...props} />;
```
vcarl, line 158: Hmmm I'm not sure we should wholesale override the click behavior for `BasicLink`. I see there's [a special case for hash links](https://github.com/stellar/new-docs/pull/169/files?diff=split&w=1#diff-94b0c1120e76c59fec00fd506741f52bR51), maybe something needs to change there?

src/constants/docsComponentMapping.js
```diff
@@ -62,7 +62,17 @@ export const components = {
...
+    a {
+      color: ${({ theme }) => theme.text};
+      @media (${MEDIA_QUERIES.canHover}) {
+        &:hover {
+          color: ${({ theme }) => theme.darkGrey};
+        }
+      }
+    }
+  `,
+  h4: styled(TextComponents.LinkedH4)`
```
vcarl, line 75: I thought we'd [deliberately stopped linking h3/h4s](https://github.com/stellar/new-docs/pull/92/files#diff-52089f61bc4808567aee5c5950816cdb)? Change looks good just want to make sure it's behavior we want

src/basics/Links.js
jeesunikim: Gotcha! I thought `LINK_DESTINATIONS.hash` would be for hash links. I like adding it to `makeLinkedHeader`!

src/constants/docsComponentMapping.js
jeesunikim: I noticed on [glossary - fee bumps](https://developers.stellar.org/docs/glossary/fee-bumps/) that [Replace by Fee](https://developers.stellar.org/docs/glossary/fee-bumps/#replace-by-fee) which is H4 is being linked to. There were links for `<H3/>` as well.

vcarl: Okay sweet! That makes sense

src/basics/Links.js
jeesunikim: I'm going to update Gatsby and see if it fixes

jeesunikim: Some updates - hash anchor tags work fine in react components. Just the ones that are addressed in .mdx files seem to be broken.

jeesunikim:
[PR 175](https://github.com/stellar/new-docs/pull/175)

PR #172, opened 2020-06-24
Fix Safari and metrics emission

vcarl:
Chrome exposes a nonstandard form of the Performance API, whose behavior I relied upon. This broke the page in other browsers. This fixes that issue by using only the standard (and more complex) behavior.

There were also issues with metrics emission, with the environment variable not being passed through the build correctly, and Google analytics not being loaded correctly due to Gatsby's behavior around env vars. 

This also improves the local production build by bringing it more closely in line with how Jenkins runs builds, which should lead to fewer shipped bugs that appear okay locally.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -57,10 +57,7 @@ exports.onCreateWebpackConfig = ({ actions, plugins }) => {
       plugins.define({
         "process.env": {
           NODE_ENV: JSON.stringify(process.env.NODE_ENV),
-          IS_BETA: JSON.stringify(process.env.IS_BETA),
-          CONTEXT: JSON.stringify(process.env.CONTEXT),
-          URL: JSON.stringify(process.env.URL),
-          DEPLOY_PRIME_URL: JSON.stringify(process.env.DEPLOY_PRIME_URL),
+          AMPLITUDE_KEY: JSON.stringify(process.env.AMPLITUDE_KEY),
         },
       }),
     ],
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test": "jest",
     "build": "gatsby build",
     "check-not-master": "[ \"$(git rev-parse --abbrev-ref HEAD)\" = \"master\" ] && echo \"\\033[31mDon't commit to master\" && exit 1 || exit 0",
-    "prod:build": "docker image build -t new-docs:localbuild .",
+    "prod:build": "cross-env TAG=new-docs:localbuild AMPLITUDE_KEY=$AMPLITUDE_KEY make docker-build",
     "prod:serve": "docker run -p 8000:80 new-docs:localbuild",
     "production": "yarn prod:build && yarn prod:serve"
   },
@@ -42,6 +42,7 @@
     "babel-preset-gatsby": "^0.4.2",
     "concurrently": "^4.1.0",
     "core-js": "2",
+    "cross-env": "^7.0.2",
     "dotenv": "^7.0.0",
     "eslint": "6.x",
     "eslint-config-prettier": "^6.9.0",
--- a/src/components/layout/LayoutBase.js
+++ b/src/components/layout/LayoutBase.js
@@ -42,9 +42,9 @@ export const LayoutBase = ({
   children,
   viewport = "width=device-width, initial-scale=1",
 }) => {
-  mark(PERFORMANCE_MARKS.layout);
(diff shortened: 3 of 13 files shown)


gatsby-node.js
```diff
@@ -57,10 +57,7 @@ exports.onCreateWebpackConfig = ({ actions, plugins }) => {
       plugins.define({
         "process.env": {
           NODE_ENV: JSON.stringify(process.env.NODE_ENV),
-          IS_BETA: JSON.stringify(process.env.IS_BETA),
-          CONTEXT: JSON.stringify(process.env.CONTEXT),
-          URL: JSON.stringify(process.env.URL),
-          DEPLOY_PRIME_URL: JSON.stringify(process.env.DEPLOY_PRIME_URL),
+          AMPLITUDE_KEY: JSON.stringify(process.env.AMPLITUDE_KEY),
```
vcarl, line 60: Gatsby has some tricky behavior with environment variables, so it's simplest just to use them directly where needed. I wanted to keep them isolated to `constants/env.js`, but I ran into issues with that. Using them directly also allows dead code elimination to slightly reduce the size of the bundle, so that's nice.

package.json
```diff
@@ -8,7 +8,7 @@
     "test": "jest",
     "build": "gatsby build",
     "check-not-master": "[ \"$(git rev-parse --abbrev-ref HEAD)\" = \"master\" ] && echo \"\\033[31mDon't commit to master\" && exit 1 || exit 0",
-    "prod:build": "docker image build -t new-docs:localbuild .",
+    "prod:build": "cross-env TAG=new-docs:localbuild AMPLITUDE_KEY=$AMPLITUDE_KEY make docker-build",
```
vcarl, line 11: This brings `yarn production` very closely in line with how Jenkins runs a build. Lots of tools (including make, docker, and gatsby) clear out environment variables for security reasons, so this removes another mismatch and thus a potential source of production-only bugs.

It passes through an `AMPLITUDE_KEY` env var for ease of testing. Otherwise, it'd have to be a package.json modification which invalidates docker cache for installing dependencies—not ideal.

src/components/layout/LayoutBase.js
```diff
@@ -42,9 +42,9 @@ export const LayoutBase = ({
   children,
   viewport = "width=device-width, initial-scale=1",
 }) => {
-  mark(PERFORMANCE_MARKS.layout);
+  mark(PERFORMANCE_MARKS.renderLayout);
   React.useEffect(() => {
-    const timing = measure(PERFORMANCE_MARKS.layout);
+    const timing = measure(PERFORMANCE_MARKS.renderLayout);
```
vcarl, line 47: 🤦

src/helpers/performance.js
```diff
@@ -6,11 +6,18 @@ export const hasHighPrecision = Boolean(performance);
...
 let internalMark = performance?.mark.bind(performance);
 let internalMeasure = (perfMark) => {
-  const { duration } = performance.measure(null, perfMark);
+  const endMark = `${perfMark} end`;
+  const measureName = `${perfMark} measure`;
+  performance.mark(endMark);
+  performance.measure(measureName, perfMark, endMark);
+  const results = performance.getEntriesByName(measureName);
+  // Get the duration off the last (most recent measurement)
+  const { duration } = results[results.length - 1];
```
vcarl, line 20: Very annoyingly, this is what's needed to get cross-browser behavior working. I tried a couple of simpler implementations, but they all had `0` durations—completely useless, as that's the only thing this is intended to measure.

src/html.js
```diff
@@ -1,15 +1,15 @@
...
+import { IS_PRODUCTION } from "constants/env";
 
 export default function HTML(props) {
   return (
     <html {...props.htmlAttributes}>
       <head>
         <meta charSet="utf-8" />
         <meta httpEquiv="x-ua-compatible" content="ie=edge" />
-        {IS_BUILD && (
+        {IS_PRODUCTION && (
```
vcarl, line 12: `IS_BUILD` relied on an environment variable that isn't consistently passed through. Gatsby used it on the server, but not on the client—so our analytics didn't load.

src/helpers/performance.js
piyalbasu, line 20: any need to worry about results being an empty array?

vcarl: We shouldn't, since it always generates a performance measure as part of the function call

Issue #170, opened 2020-06-23
Right side navigation stops working once a user clicks H2 linked icon


jeesunikim:
![bug-right-side-nav-linked-header](https://user-images.githubusercontent.com/3912060/85419917-9b85e800-b540-11ea-9c02-81f0f9ec4354.gif)

### What did you do?
Go to [/docs](https://developers.stellar.org/docs/) that has the right side navigation and scroll down the page.

### What did you expect to see?
The right side navigation to highlight correctly as I scroll down each section on the page after clicking the clip icon that I see when hovering `H2` tags.

### What did you see instead?
The right side navigation stopped highlighting the section that is visible on the page after clicking the clip icon.

jeesunikim:
[Comment](https://github.com/stellar/new-docs/pull/175#issuecomment-651290390) regarding to this to keep track of it

shanzzam:
This is still an issue so am adding it to the Nebula team backlog as it's a technical issue.

PR #175, opened 2020-06-29
Updated gatsby and added a function in gatsby-browser.js to enable scroll to hashtag

jeesunikim:
**Problem:**
hash-links-bug-480

On /docs' [Run Core Node - configuring](https://developers.stellar.org/docs/run-core-node/configuring/), for example, if you click links that are targeting hash links. They won't work most of the time (some of the hash links I clicked for the first time worked; I've tried multiple times on incognito, but happens randomly).

Gatsby's [Recommendations for programmatic, in-app navigation](https://www.gatsbyjs.org/docs/gatsby-link/#recommendations-for-programmatic-in-app-navigation) recommends to use an `anchor tag` (which what we were already using) or `reach/router`'s `navigate`.

**UPDATING GATSBY AND ITS DEPENDENCIES**
I thought maybe updating gatsby would fix scrolling issues but it did not! But I think it's good to keep our dependencies updated so making a PR of it anyway.

I did `yarn outdated` and installed to the version under `Wanted` - double checked locally, didn't see any issues.

yarn-outdated

**Scroll to a hash tag section when anchor hash tag is clicked**

- Anchor hash tag doesn't seem to work in mdx. I played with plain `<h2/>` tag and it didn't work. I couldn't figure out why even by reading MDX documentation. I decided to use `gatsby-browser.js`'s `shouldUpdateScroll` function to target the section
- I thought of using a library such as “gatsby-remark-autolink-headers”, but this one is using for older version - it's using react router not reach router


vcarl:
I wonder if this is the same root issue as #170 , I notice that on a totally fresh tab, opening e.g. https://developers-pr175.previews.kube001.services.stellar-ops.com/docs/run-core-node/configuring/#example-configurations works correctly, and I'm able to click a heading and have it work correctly once. The right side nav + header links stop working at the same time.

This is a solid immediate fix for the behavior, but we might be able to remove this logic upon fixing #170.

PR #179, opened 2020-07-06
Streamline shared components, move files to clearer folders

vcarl:
This started as fixing a visual inconsistency with single API reference pages,

Screen Shot 2020-07-06 at 1 33 14 PM
Screen Shot 2020-07-06 at 1 33 19 PM

but as I fixed that I realized there were a number of small unexpected homes for components. This more strongly separates components by api reference/documentation, and consolidates some cruft that we accumulated.


README.md
```diff
@@ -24,7 +24,7 @@
     - [`<Endpoint/>`](#endpoint)
     - [`<EndpointsTable/>`](#endpointstable)
     - [`<ExampleResponse/>`](#example-json-response)
-    - [`<NavTable/>`](#navtable)
+    - [`<MethodTable/>`](#MethodTable)
```
vcarl, line 27: `NavTable` was just a thin wrapper over `MethodTable`, so I removed the extra component.

src/components/ApiReference/NavItem.js
```diff
@@ -0,0 +1,105 @@
...
+    props.isActive
+      ? css`
+          ${activeStyles}
+        `
+      : ""}
+`;
+
+const NAV_BAR_HEIGHT = 89;
+const FIXED_NAV_DISTANCE = 140 + NAV_BAR_HEIGHT;
+export const NavItem = ({ isActive, forwardedRef, children, depth }) => {
```
vcarl, line 41: I peel this out of `templates/ApiReference.js` both to unify styles on the single/full API reference pages, and because the template was fairly hard to navigate. Template had a lot of things all living in one place.

src/components/ApiReference/ScrollRouter.js
```diff
@@ -5,7 +5,13 @@ import throttle from "lodash/throttle";
...
 import { SideNavProgressContext } from "components/SideNav/Provider";
 
-export const Context = React.createContext();
+export const Context = React.createContext({
+  isScrollingDown: false,
+  stopTrackingElement: () => {},
+  trackElement: () => {},
+  onLinkClick: () => {},
+  setIsNavClicked: () => {},
+});
```
vcarl, line 14: I needed to add defaults here because the single API reference pages don't have the provider. The value provided here gets used if the context is consumed without a provider at the top level, so just stubs.

src/components/ApiReference/SharedStyles.js
```diff
@@ -0,0 +1,63 @@
...
+export const NestedRow = styled.div`
+  display: grid;
+  grid-template-columns: repeat(9, 4rem);
+  column-gap: calc((100% - ${9 * 4}rem) / ${9 - 1});
+
+  @media (${MEDIA_QUERIES.gtXlDesktop}) {
+    grid-template-columns: repeat(18, 4rem);
+    column-gap: calc((100% - ${18 * 4}rem) / ${18 - 1});
+  }
+`;
```
vcarl, line 63: All of these styles were in `Documentation/SharedStyles.js`, but they weren't used on any pages that used the `Documentation.js` template. I moved them here so their use is clearer.

src/components/Footer.js
```diff
@@ -6,7 +6,7 @@ import { PALETTE, FONT_WEIGHT, MEDIA_QUERIES } from "constants/styles";
 import { Link } from "basics/Links";
 import { H4, H6 } from "basics/Text";
 
-import { Subscribe } from "./Subscribe";
+import { Subscribe } from "components/Footer/Subscribe";
```
vcarl, line 9: I moved `Footer` to the `components/` folder, with `components/Footer/X.js` files for stuff that's exclusively used by the footer

src/constants/docsComponentMapping.js
```diff
@@ -124,3 +131,72 @@ export const components = {
...
+
+/**
+ * Template-specific overrides to these default styles
+ */
+
+const { td: TD } = components;
+
+export const documentationComponents = {
+  ...components,
+  wrapper: WrapperDocumentation,
```
vcarl, line 143: I moved wrappers into their own component files, and moved the template-specific component definitions to all live in here

src/helpers/makeLinkedHeader.js
```diff
@@ -41,8 +42,7 @@ export const makeLinkedHeader = (
 
   const WrappedComponent = (props) => {
     const { children } = props;
-    const id =
-      props.id || slugify(Array.isArray(children) ? children.join() : children);
+    const id = props.id || slugify(loopAndExtractString(children));
```
vcarl, line 45: We just hadn't hit this bug, but it was waiting to happen. Use the recursive string extraction helper to build this `id`

src/templates/ApiReference.js
```diff
@@ -1,207 +1,58 @@
...
-      depth={depth}
-      ref={itemRef}
-      onClick={() => {
-        setIsNavClicked(true);
-      }}
-    >
-      {children}
-    </NavItemEl>
-  );
-};
```
vcarl, line 204: One of the main goals of this PR. This and the other template are much easier to read.

src/components/WrapperApiReference.js
```diff
@@ -0,0 +1,52 @@
...
+      ),
+    [children],
+  );
+  const MiddleColumnContent = React.useMemo(
+    () =>
+      React.Children.toArray(children).filter(
+        (child) => !RIGHT_COLUMN_COMPONENTS_NAME[child.props.mdxType],
+      ),
+    [children],
+  );
```
piyalbasu, line 29: we're just splitting into 2 arrays right? 
might be a little cleaner to use 

_.partition https://lodash.com/docs/4.17.15#partition

or just plain JS to divide: https://stackoverflow.com/a/47225591

src/components/WrapperApiReference.js
```diff
@@ -0,0 +1,52 @@
...
+const RIGHT_COLUMN_COMPONENTS_NAME = {
+  CodeExample: "CodeExample",
+  EndpointsTable: "EndpointsTable",
+  ExampleResponse: "ExampleResponse",
+  MethodTable: "MethodTable",
+};
+// API reference has 2 columns, with certain types of content always being in one
+// or the other. This component sorts through its children to separate them.
+export const WrapperApiReference = ({ children, ...props }) => {
+  const rightColumnContent = React.useMemo(
```
piyalbasu, line 16: NIT: RightColumnContent to match case of MiddleColumnContent?

src/components/WrapperApiReference.js
vcarl: Oh good call, partition is def the right tool

src/components/WrapperApiReference.js
vcarl: Will update to `middleColumnContent` since it's not a component

PR #180, opened 2020-07-07
Attempt to fix SEO issues on /api/* pages

vcarl:
We added "single" pages for each section of the API Reference so that crawlers know to associate them with that URL, but it assumes that crawlers won't execute javascript. Per [Google's recommendations](https://developers.google.com/search/docs/guides/dynamic-rendering#cloaking), this adds a user agent check to redirect all crawlers to the JS-less form of the page.

HOWEVER we know that some of these crawlers execute JS, and the existing page will overwrite that single page's worth of content with the entirety of the /api/* pages, so this adds a check to only client-render a single section if `javascript=false` is present in the querystring.

It also tries to encourage crawlers to find the right content by adding multiple `<main>` tags, with a `hidden` attribute if it's not referred to by the active path.

### DIFF
--- a/src/components/ApiReference/ReferenceSection.js
+++ b/src/components/ApiReference/ReferenceSection.js
@@ -13,15 +13,46 @@ import { Link } from "basics/Links";
 import { Route } from "components/ApiReference/Route";
 import { CustomColumn, NestedRow } from "components/ApiReference/SharedStyles";
 import { TrackedContent } from "components/SideNav";
+import { useLocation } from "@reach/router";
+import { Context as ScrollRouterContext } from "./ScrollRouter";
 
 const { h1: H1, h2: H2 } = apiReferenceComponents;
 
-const SectionEl = styled.article`
+const MainEl = styled.main`
+  display: block;
+`;
+const ArticleEl = styled.article`
   display: block;
   &:first-child {
     margin-top: 5rem;
   }
 `;
+
+const Article = ({ children, path }) => {
+  // The scroll router overrules Reach Router once the page loads, but we need
+  // Reach router's location to get initial path
+  const { pathname } = useLocation();
+  const [isActive, setIsActive] = React.useState(pathname);
+  const { subscribe } = React.useContext(ScrollRouterContext);
+
+  React.useEffect(
+    () =>
+      subscribe((newPath) => {
+        setIsActive(path === newPath);
+      }),
+    [subscribe, path],
+  );
+
+  return (
+    <MainEl name={path} hidden={!isActive}>
+      <ArticleEl>{children}</ArticleEl>
+    </MainEl>
+  );
+};
+Article.propTypes = {
+  children: PropTypes.node.isRequired,
+  path: PropTypes.string.isRequired,
+};
 
 export const ReferenceSection = React.memo(
   ({ body, relativePath, title, githubLink, path }) => {
@@ -39,7 +70,7 @@ export const ReferenceSection = React.memo(
     );
 
     return (
-      <SectionEl>
+      <Article path={path}>
         <Route originalFilePath={relativePath} path={path}>
           <TrackedContent identifier={path}>
             <NestedRow>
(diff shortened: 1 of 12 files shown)


src/components/ApiReference/ScrollRouter.js
```diff
@@ -60,7 +60,11 @@ export const ScrollRouter = ({ children, initialActive = "" }) => {
       // before, update the route.
       if (newActiveNode && newActiveNode !== activeNode) {
         setActiveNode(newActiveNode);
-        window.history.replaceState(null, null, routeMap.get(newActiveNode));
+        window.history.replaceState(
+          null,
+          null,
+          `${routeMap.get(newActiveNode)}${window.location.search}`,
```
piyalbasu, line 66: should we be concerned about google indexing the url w the query param? not sure if they do that. 

for ex: https://developers.stellar.org/api/introduction/pagination/?javascript=false

vcarl: hmmm this is an excellent question

src/components/ApiReference/ReferenceSection.js
```diff
@@ -0,0 +1,105 @@
...
+  React.useEffect(
+    () =>
+      subscribe((newPath) => {
+        setIsActive(path === newPath);
+      }),
+    [subscribe, path],
+  );
+
+  return (
+    <MainEl name={path} hidden={!isActive}>
```
vcarl, line 47: I added a `<main>` element wrapping each section. There's only supposed to be 1 main element on a page, so it also adds a `hidden` attribute if the current section doesn't match the current URL. And this is maybe a bit of magical thinking, but it also adds the current URL as the `name` attr, cuz hey maybe that'll help?

src/components/ApiReference/ScrollRouter.js
```diff
@@ -34,10 +37,14 @@ export const ScrollRouter = ({ children, initialActive = "" }) => {
   const isScrollingDown = React.useRef(false);
 
   // Navigation
-  const onLinkClick = React.useCallback(function onLinkClick(route) {
-    window.history.pushState(null, null, route);
-    smoothScrollTo(elementMap.get(route).current, { duration: 0 });
-  }, []);
+  const onLinkClick = React.useCallback(
+    function onLinkClick(route) {
+      emit(route);
```
vcarl, line 42: There's no mechanism to watch for URL changes on a page level, so I updated the scroll router to support subscriptions. Any time the route updates, all subscribers will be notified.

src/components/layout/Seo.js
```diff
@@ -12,7 +12,14 @@ import favicon192 from "assets/favicon/android-chrome-192x192.png";
...
 import StellarDocsImage from "assets/images/og_developers.jpg";
 
-export const Seo = ({ title, description = "", path }) => {
+export const Seo = ({
+  title,
+  description = "",
+  path,
+  omitPredicate = () => true,
+  meta = [],
+  link = [],
```
vcarl, line 21: Arguably overengineered. We want to omit the `<link rel="canonical">` head tag on the `/api/` page (because it's not the canonical URL for any of that content and Google is confused), but I didn't want to special-case it

src/helpers/useSubscription.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+/**
+ * useSubscription makes it easy to make a simple event emitter.
+ * @returns {object} An object with functions `subscribe` and `emit`. The
+ * `subscribe` method accepts a callback function and returns an `unsubscribe`
+ * function, and `emit` will broadcast a value to all subscribers
+ */
+export const useSubscription = () => {
```
vcarl, line 9: I lowkey love little subscriber patterns like this.
```js
const unsubscribe = subscribe(newValue => doSomethingWith(newValue))
```
This is a handy dandy little hook for quickly creating them.

src/components/ApiReference/ScrollRouter.js
vcarl: I _believe_ this will work fine, because we have a canonical URL set. Google appears to have retired all the tools it used to expose for manually mucking with the URL params in their webmaster tools

PR #183, opened 2020-07-09
Fix mismatched JS-less render

vcarl:
Single API reference pages (loaded with `?javscript=false` were visually broken due to a server/client mismatch. This corrects that to ensure the content is identical, but I wish Gatsby would just trash the HTML in the event of a mismatch!


jeesunikim:
I noticed a huge gap above all H1, @vcarl 
h1-space

vcarl:
Yeah, I think the preview link is out of date, I fixed that in the most recent commit. Having some cache issues now that they're using a per-PR url
Screen Shot 2020-07-10 at 10 40 03 AM

PR #187, opened 2020-07-13
Add 301 redirect on /no-js pages

vcarl:
Final (🤞) SEO adjustment tweak. Search results are starting to point to the right material, but some of the results include the `/no-js` segment, which is intended to be invisible. The 301 should help direct to the right place.

### DIFF
--- a/src/constants/docsComponentMapping.js
+++ b/src/constants/docsComponentMapping.js
@@ -179,12 +179,8 @@ const OrangeTableCell = styled.td`
 export const apiReferenceComponents = {
   ...components,
   wrapper: WrapperApiReference,
-  h1: styled(components.h1).attrs(() => ({
-    as: makeLinkedHeader(ApiRefH1, headerOptions),
-  }))``,
-  h2: styled(components.h2).attrs(() => ({
-    as: makeLinkedHeader(ApiRefH2, headerOptions),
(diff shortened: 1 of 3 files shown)


vcarl:
Actually, I just noticed the canonical URL isn't right on specific sub-pages.

src/constants/docsComponentMapping.js
```diff
@@ -179,12 +179,8 @@ const OrangeTableCell = styled.td`
...
   ...components,
   wrapper: WrapperApiReference,
-  h1: styled(components.h1).attrs(() => ({
-    as: makeLinkedHeader(ApiRefH1, headerOptions),
-  }))``,
-  h2: styled(components.h2).attrs(() => ({
-    as: makeLinkedHeader(ApiRefH2, headerOptions),
-  }))``,
+  h1: makeLinkedHeader(ApiRefH1, headerOptions),
+  h2: makeLinkedHeader(ApiRefH2, headerOptions),
```
vcarl, line 183: Styled components' `as` prop get wibbly when there are multiple levels of components in question. Since there aren't any additional styles here, I think this is complexity with no benefit, and this appears to be how some unexpected styles are being added to the static HTML (though it's weird that it's not also applying to the client render)

PR #190, opened 2020-07-14
Fix redundant scroll, weird header margin on /docs/*

vcarl:
Scrolling to the right section is handled [when the page initially loads](https://github.com/stellar/new-docs/blob/daa27caedecb0da30b9b4e0b4b17d3243ad1d962/src/html.js#L27-L32), this logic is redundant and causes content to scroll twice, which is a little disruptive. The downside of removing this is that it won't scroll when the page is loaded in dev mode (`yarn start`) because Gatsby doesn't serve any HTML in that case.

Headers before/after:
Screen Shot 2020-07-14 at 2 20 01 PM
Screen Shot 2020-07-14 at 2 20 05 PM

Loading scroll before/after:
Before: https://gfycat.com/shrillregularaardvark
After: https://gfycat.com/energeticverifiableewe

### DIFF
--- a/src/basics/Text.js
+++ b/src/basics/Text.js
@@ -54,7 +54,6 @@ export const H1 = styled.h1`
   display: inline-block;
   font-size: 2rem;
   margin-bottom: 0.5rem;
-  margin-top: 5.25rem;
   line-height: 1.25;
   letter-spacing: normal;
 `;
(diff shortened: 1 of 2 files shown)


vcarl:
The header changes position twice in the "before" video, only once in the "after"

PR #192, opened 2020-07-15
Fix redirect loop for crawlers

vcarl:
Unfortunately this leaves /no-JS pages accessible by directly navigating to them, which isn't what we want. I can't manage to get nginx to do a rewrite for crawlers and a redirect for visitor, though, it always ends in either a 404 (because it stopped looking for fields after the rewrite) or a redirect loop.

I've been flying a little blind so far, but dev tools let's you change the user agent to simulate crawler requests, so I'm more confident that this is going to be what we need.

### DIFF
--- a/nginx/includes/server.conf
+++ b/nginx/includes/server.conf
@@ -6,31 +6,36 @@ location = / {
   proxy_ssl_server_name on;
 }
 
-location ~ ^/no-js/(.*)$ {
-  # `$scheme://$http_host` is because otherwise, it adds the port nginx is
-  # listening on. Since we listen on 80 but expose 8000 by default from docker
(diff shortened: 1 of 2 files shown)


nginx/includes/server.conf
```diff
@@ -6,31 +6,36 @@ location = / {
...
+  set $noJs false;
+  set $needsCrawlerRedirect '';
+
+  if ($arg_javascript = 'false') {
+    set $noJs true;
+  }
+  if ($noJs = true) {
+    rewrite ^/api/(.*) /no-js/api/$1;
+  }
+  if ($noJs = false) {
```
piyalbasu, line 25: oof `if`'s in nginx are brutal

PR #203, opened 2020-07-26
Add Python implementation of the code example. (issue asset)

overcat:


RONIZAM80:
Add python implement 203

PR #225, opened 2020-09-02
Update decentralized-exchange.mdx

phil-stellar:
Trading terminology will be confusing because "offers" is typically associated with "sell orders" and synonymous with "asks." I know the terminology in the code is "offers" and changing "offers" to "orders" here might be confusing.

(minor comment): Using "consumed" instead of "taken" because when you "take an offer" you are typically buying an ask (alternatively, selling a bid is "hitting a bid")...just conventional trading lingo

"Prices are specified as a `{numerator, denominator}` pair, with both components of the fraction represented as 32 bit signed integers. When expressing a price of "Asset A in terms of Asset B", the amount of B is always the numerator, and A is always the denominator. "
 - this is backwards...with currency pairs, "Asset A in terms of Asset B" usually means Asset B is the quote asset, and A is the base asset. These are always quoted as base/quote (so A is numerator and B is denominator). When you're talking about buying/selling an fx pair, you're generally talking in base terms (i.e. I'm buying/selling yen... which would be represented yen/(quoteasset).

### DIFF
--- a/content/docs/glossary/decentralized-exchange.mdx
+++ b/content/docs/glossary/decentralized-exchange.mdx
@@ -64,15 +64,15 @@ Looking at the diagram, there are a few orientation things to notice and think a
 
 Some assets will have a very thin or nonexistent orderbook between them. That's fine: as discussed in greater detail below, paths of orders can facilitate exchange between two thinly traded assets.
 
-## Cross-asset payments (in fx this is referred to as multi-leg transactions)
+## Cross-asset payments 
 
-Suppose you are holding sheep and want to buy something from a store that only accepts wheat. You can create a payment in Stellar that will automatically convert your sheep into wheat. It goes through the sheep/wheat orderbook and converts your sheep at the best available rate.
+Suppose you are holding sheep and want to buy something from a store that only accepts wheat. You can create a payment in Stellar that will automatically convert your sheep into wheat. It goes through the sheep/wheat orderbook and converts your sheep at the best available rate. 
 
 You can also make more complicated paths of asset conversion. Imagine that the sheep/wheat orderbook has a very large spread or is nonexistent. In this case, you might get a better rate if you first trade your sheep for brick and then sell that brick for wheat. So a potential path would be 2 hops: sheep->brick->wheat. This path would take you through the sheep/brick orderbook and then the brick/wheat orderbook.
 
 These paths of asset conversion can contain up to 6 hops, but the whole payment is atomic--it will either succeed or fail. The payment sender will never be left holding an unwanted asset.
 
-This process of finding the best path of a payment is called **pathfinding**. Pathfinding involves looking at the current orderbooks and finding which series of conversions gives you the best rate. It is handled outside of Stellar Core by something like Horizon.
+This process of finding the best path of a payment is called **pathfinding**. Pathfinding involves looking at the current orderbooks and finding which series of conversions gives you the best rate. It is handled outside of Stellar Core by something like Horizon. In foreign exchange this is often referred to as multi-leg and cross-currency transactions.
 
 ## Preferred currency
 
--- a/content/docs/glossary/decentralized-exchange.mdx
+++ b/content/docs/glossary/decentralized-exchange.mdx
@@ -3,41 +3,42 @@ title: Decentralized Exchange
 order:
 ---
 
-In addition to supporting the issuing and movement of [assets](./assets.mdx), the Stellar network also acts as a decentralized **distributed exchange** that allows you to trade and asset on the network. The Stellar ledger stores both balances held by user accounts and offers that user accounts make to buy or sell assets.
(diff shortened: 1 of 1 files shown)


content/docs/glossary/decentralized-exchange.mdx
```diff
@@ -3,41 +3,42 @@ title: Decentralized Exchange
...
+It's important to note that the price you set is unrelated to the fee you pay for submitting the order as a part of a transaction. Fees are always paid in the native currency of the network (lumens), and are related to the transaction that you submit to the network (which contains your order operation) as opposed to your order itself.
 
 For more information, take a look at [our guide on fees in Stellar](./fees.mdx).
 
-## Passive Offers
+## Passive Order
 
-**Passive offers** allow markets to have zero spread. If you want to offer USD from anchor A for USD from anchor B at a 1:1 price, you can create two passive offers so the two offers don't fill each other.
+**Passive orders** allow markets to have zero spread. If you want to exchange USD from anchor A for USD from anchor B at a 1:1 price, you can create two passive orders so the two orders don't fill each other. (will this actually be a zero spread? it should still be a tick wide right? or would it be a "locked market", where the bid and the ask is at the same price?)
```
debnil, line 37: Flagging the last sentence. It looks like a comment, rather than an edit. We should make sure to resolve this question and not push that to `master`.

I don't know the answer, but it seems like it could be any of the above depending on the orders. The documentation wording seems deliberately vague, to just illustrate that this is a possibility - rather than saying "this is how you do it."

content/docs/glossary/decentralized-exchange.mdx
```diff
@@ -3,41 +3,42 @@ title: Decentralized Exchange
...
 
-A passive offer is an offer that does not take a counteroffer of equal price. It will only fill if the prices are not equal. For example, if the best offer to buy BTC for XLM has a price of 100XLM/BTC, and you make a passive offer to sell BTC at 100XLM/BTC, your passive offer _does not_ take that existing offer. If you instead make a passive offer to sell BTC at 99XLM/BTC it would cross the existing offer and fill at 100XLM/BTC.
+A passive order is an order that does not execute against a marketable counter order with the same price. It will only fill if the prices are not equal. For example, if the best order to buy BTC for XLM has a price of 100XLM/BTC, and you make a passive offer to sell BTC at 100XLM/BTC, your passive offer _does not_ take that existing offer. If you instead make a passive offer to sell BTC at 99XLM/BTC it would cross the existing offer and fill at 100XLM/BTC. (i have some questions here...why does it not execute at a locked market (i.e. equal price) but at a crossed market it will execute?
```
debnil, line 39: Flagging this last sentence too, so we don't push a question. I don't know the details of this, but I think either @nikhilsaraf or @jonjove will know.

content/docs/glossary/decentralized-exchange.mdx
```diff
@@ -63,7 +64,7 @@ Looking at the diagram, there are a few orientation things to notice and think a
 
 Some assets will have a very thin or nonexistent orderbook between them. That's fine: as discussed in greater detail below, paths of orders can facilitate exchange between two thinly traded assets.
 
-## Cross-asset payments
+## Cross-asset payments (in fx this is referred to as multi-leg transactions)
```
debnil, line 67: This parenthetical should be added to the below paragraph, rather than the title.

content/docs/glossary/decentralized-exchange.mdx
phil-stellar: if this is similar to maker/taker orders it should read something like this:

A passive order is a limit order that does not immediately execute against a marketable counter order on the orderbook. It adds liquidity to the orderbook and will only be filled if another order non-passive order trades against it. For example, if there is an ask to sell 1 XLM at $0.10, and you create a passive buy order to buy to buy 1 XLM for $0.10 (or above), your passive buy order _does not_ take that existing $0.10 ask. Instead, it will place your bid at the highest possible price up to your limit price without executing the order, in this case ($0.09999999). It is a passive order because it will never initiate an order execution. It is an order that adds liquidity to and another order has to be initiated against it for it to execute.

content/docs/glossary/decentralized-exchange.mdx
phil-stellar: Let's clarify this...see the comment for line #39... I'm not sure what happens with locked/crossed markets in Stellar. Might be easier to hop on call with anyone else necessary to go over this.

jonjove: If you create two passive offers at the same price, it will actually be zero spread. This is the only way to get a zero spread market in Stellar. (I'd need to double check, but it might also be possible to get a zero spread market by modifying an offer that was created passive. Either way, this is conceptually the same behavior.)

content/docs/glossary/decentralized-exchange.mdx
jonjove: @phil-stellar This isn't the same as maker/taker orders. The behavior of create passive offer is not post only.

rice2000 (approved):
Thought this was already merged.  Merging now!

PR #238, opened 2020-09-18
Update Send a Payment golang example to use correct method

definitepotato:
**Summary**
For the example code of "Send a Payment" in the main() function lines 16 and 24 are not using the correct method, this should be client.AccountDetail() not client.Accounts()

**Explanation**
```
15:  destAccountRequest := horizonclient.AccountRequest{AccountID: destination}
16:  destinationAccount, err := client.Accounts(destAccountRequest)
17:  if err != nil { 
18:      panic(err)
19:  }
```

I'll use one of the two code blocks since they both have the same issue.  In the code block above line 15 creates an endpoint for a single account using the horizonclient.AccountRequest() method.  The following line 16 then sends a query to fetch accounts who have a signer.  Problem is this is an endpoint for multiple accounts, not a single account.  line 15 creates an endpoint for a single account, so line 16 is expecting horizonclient.AccountsRequest() not horizonclient.AccountRequest.

**Solution**
Change line 16 to call client.AccountDetail() rather than client.Accounts()

### DIFF
--- a/content/docs/tutorials/send-and-receive-payments.mdx
+++ b/content/docs/tutorials/send-and-receive-payments.mdx
@@ -140,7 +140,7 @@ func main () {
 
     // Make sure destination account exists
     destAccountRequest := horizonclient.AccountRequest{AccountID: destination}
-    destinationAccount, err := client.Accounts(destAccountRequest)
+    destinationAccount, err := client.AccountData(destAccountRequest)
     if err != nil {
         panic(err)
     }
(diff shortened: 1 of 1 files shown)


tyvdh:
@rice2000 Who's on task to review golang api docs?

rice2000:
Good question.  I'm not sure.  Is it out of reach for you @tyvdh?

vcarl:
@Shaptic maybe? Last person at SDF to submit a Go example change

Shaptic:
The Horizon team is prioritizing Protocol 14 testing right now, but yes, I can review Go-related SDK PRs once that's over.

@definitepotato Can you also update the partial snippets lower on [that page](https://developers.stellar.org/docs/tutorials/send-and-receive-payments/#send-a-payment)? This change currently only updates the large example, and they'd be out of sync.

definitepotato:
@Shaptic absolutely, I'll go ahead and do that now.

Shaptic (approved):
@rice2000 @vcarl I don't have write access, but this LGTM. :+1: Thanks @definitepotato!

rice2000 (approved):
Fantastic!  Great work @definitepotato.  And thanks for looking it over, @Shaptic.  Team work. Dream work.

PR #241, opened 2020-09-18
Added notes on issuer vs. distribution accounts

### LINKED ISSUE #240: Add notes on the issuer and distribution account pattern
Distributing assets through a distribution account is a design pattern. Functionally - you can do away with the distribution account and distribute directly from the issuer account

A less know fact is that you can even create a market directly from the issuing account and issue by trading.

With that said, there are two main reasons (that I'm aware of) to use a distribution account

First and foremost is security.

The account you distribute from is going to be a "hot" account. Meaning that some web service out there has direct access to sign distribution transactions.

If that gets compromised and it's an issuing account - whoever took control can now print as much of your asset as they want. that's pretty bad.

If that gets compromised and it's a distribution account with a limited amount then the stakes are lower. And once discovered the issuer account can essentially revoke the compromised account and start fresh with a new one, without changing anything about the issuer account.

The other reason is book keeping. The issuer account can't hold a balance of their own asset which makes book keeping a hassle. If you allocate an inventory of your own asset then it's easier to track money. This is pretty common in various ledgering solutions.

And, as an added bonus, distribution accounts decouple our ecosystem standards from issuance. This allows us to come up with crazy concepts like non-issuing anchors without actually changing protocols.

JakeUrban:
resolves #240

### DIFF
--- a/content/docs/issuing-assets/how-to-issue-an-asset.mdx
+++ b/content/docs/issuing-assets/how-to-issue-an-asset.mdx
@@ -12,9 +12,11 @@ It’s a pretty simple process that requires four operations: one to create an i
 
 <Alert>
 
+
 Note: you don't actually have to issue assets to a dedicated distribution account: you can issue them to any account with the requisite trustline. However, using a distribution account is the recommended practice, and it makes the process a lot easier to explain. So that's how we'll do it in this guide.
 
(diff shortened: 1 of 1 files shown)


rice2000 (approved):
Wowsers.

PR #255, opened 2020-10-07
Linked heading styles fixed

quietbits:
The issue was with nested styled-component in `makeLinkedHeader` helper. Moving link style to a class in `GlobalStyles` fixed the issue. It's not ideal but works for this edge case.

**Before**
image

**After**

image


src/basics/GlobalStyles.js
```diff
@@ -1,11 +1,32 @@
...
-import { createGlobalStyle } from "styled-components";
+import { createGlobalStyle, css } from "styled-components";
 import styledNormalize from "styled-normalize";
 
 import { FONTS } from "constants/fonts";
-import { FONT_FAMILY } from "constants/styles";
+import { FONT_FAMILY, MEDIA_QUERIES } from "constants/styles";
 import { expansionStyles } from "components/Expansion";
 
+const linkedHeadingStyles = css`
```
vcarl, line 9: I think a comment explaining why this exception is needed (to the best of our knowledge anyway haha) would be good to have here

quietbits: Good point, PR comment is not enough. :+1:

vcarl (approved):
🎉

PR #259, opened 2020-10-20
Rice2000 remove account lock

rice2000:
There are 2 errant spots in the doc that suggest locking an account by setting the master key to 0.  This isn't best practice at this point, and causes unnecessary heartache, so this PR gets rid of them.


content/docs/glossary/multisig.mdx
```diff
@@ -143,50 +143,4 @@ Expense account setup:
...
 </CodeExample>
 
-<Alert>
 
-The next two examples involve setting the master key weight of an account to 0. Be very, very careful if you decide to do that: that key will no longer be able to sign any kind of transaction, so you are in danger of _permanently_ locking yourself out of your account. Make sure you've thought carefully about what you're doing, and that you understand the implications.
-
-</Alert>
-
-### Example 4: Company Accounts
-
```
leighmcculloch, line 153: The company account example might still be useful. It is not encouraging the lock accounts approach so I think it can stay?

leighmcculloch (approved):
Looks good, but I think we can keep example 4.

content/docs/glossary/multisig.mdx
rice2000: That's a fair point.  I added a bit to the warning + the example to notify people to be careful about the order they follow when changing weights.  Thoughts @leighmcculloch ?

content/docs/glossary/multisig.mdx
```diff
@@ -158,7 +158,7 @@ Company account setup:
 <CodeExample>
 
 ```
-  Master Key Weight: 0 (Turned off so this account can't do anything without an employee)
+  Master Key Weight: 0 (Turned off so this account can't do anything without an employee.  Make sure to set this weight *last* so you don't lock the account before setting up the other signers.)
```
leighmcculloch, line 161: It actually doesn't need to be the last, it can be in the transaction that adds other signers. I'm not sure this comment is necessary.

leighmcculloch: But you're correct that you can't set the master weight to zero then in a separate transaction go ahead and add a signer.

rice2000: You're right.  I think the Alert above does enough to warn people, so it's fine without this line.  I'm getting rid of it, then merging.  Thanks for your input!

PR #264, opened 2020-10-27
Update allow-trust.mdx

rice2000:
Change trust is now an int, not a boolean.  That's because of the Protocol 13 "fine-grained asset auth" adddition.

### DIFF
--- a/content/api/resources/operations/object/allow-trust.mdx
+++ b/content/api/resources/operations/object/allow-trust.mdx
@@ -26,7 +26,7 @@ See the [`Allow Trust` errors](../../../errors/result-codes/operation-specific/a
   - The code for the asset.
 - authorize
   - int
-  - Flag indicating whether the trustline is authorized.  1 if the account is authorized to transact with the asset. 2 if the account is authorized to maintain orders, but not to perform other transactions. 0 if the account is not authorized to transact with the asset in any way.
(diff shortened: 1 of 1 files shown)


content/api/resources/operations/object/allow-trust.mdx
```diff
@@ -26,7 +26,7 @@ See the [`Allow Trust` errors](../../../errors/result-codes/operation-specific/a
   - The code for the asset.
 - authorize
   - boolean
```
Shaptic, line 28: `boolean` -> `int`, right?

rice2000: Fixed!  Other than that, does it check out, @Shaptic?

Shaptic: I'd personally prefer to put the possible values of the flag in order, but I do see merit in a "sorted by order of interest" approach, as well. In either case, LGTM! However, I will say I can't actually confirm the _content_ of the docs (that is, I don't know if those numbers for `authorize` match reality).

Shaptic: (I initially just hopped into this PR for my own edification.)

rice2000: That's a fair point re: the ordering.  I tinkered with it a bit, tried it one way, then another.  Now I'm reverting back to what I started with, which is also what you're suggesting.

jeesunikim (approved):

PR #266, opened 2020-11-02
Update index.md for with withdrawal directory.

greed2411:
hey just fixed the directory ending name to `withdrawal` from `withdraw` to fix the generated link in the new docs. let me know if there is anything else.


greed2411:
closing it since jrice already has a PR doing the same.

PR #274, opened 2020-11-11
Update Makefile to label image with timestamp

jacekn:
The label can be used to determine build timestamp which can be
different from image tag or layer push timestamp


jcooter (approved):
LGTM

brahman81 (approved):
LGTM

PR #275, opened 2020-11-16
Improve font loading

quietbits:
- Move fonts to CSS file and load it in `head` (for better caching).
- Remove "Helvetica Neue" font from fallbacks because it's a different style (serif instead of sans-serif).
- Set default font in CSS file. This way, if JavaScript is disabled in the browser, the font will be set correctly.


Ogungodspower5 (approved):

vcarl (approved):
This looks good! Unfortunately the PR previews are set to `no-cache` to avoid stale previews, so tricky to confirm that behavior.

PR #290, opened 2020-12-05
Expand the Claimable Balance page to include JS example code.

Shaptic:
Relevant *live* page is [here](https://developers.stellar.org/docs/glossary/claimable-balance/#example).

### DIFF
--- a/content/docs/glossary/claimable-balance.mdx
+++ b/content/docs/glossary/claimable-balance.mdx
@@ -97,7 +97,7 @@ async function main() {
     .build();
 
   tx.sign(A);
-  return await server.submitTransaction(tx).then(function() {
+  await server.submitTransaction(tx).then(function() {
     console.log("Claimable balance created!");
   }).catch(function (err) {
     console.error(`Tx submission failed: ${err}`)
@@ -258,7 +258,7 @@ tx = new sdk.TransactionBuilder(aAccount, {fee: sdk.BASE_FEE})
   .build();
 
 tx.sign(A);
-return await server.submitTransaction(tx).catch(function (err) {
+await server.submitTransaction(tx).catch(function (err) {
   console.error(`Tx submission failed: ${err}`)
 });
 ```
--- a/content/docs/glossary/claimable-balance.mdx
+++ b/content/docs/glossary/claimable-balance.mdx
@@ -56,7 +56,7 @@ Each of these accounts can only claim the balance under certain, individual cond
 <CodeExample>
 
 ```js
-const sdk = require("stellar-sdk");
+const sdk = require("stellar-sdk")
 
 var server = new sdk.Server("https://horizon-testnet.stellar.org")
 
@@ -96,69 +96,6 @@ server
     tx.sign(A)
     return [tx, server.submitTransaction(tx)]
   })
-  // Method 1: Suppose `tx` comes from the transaction built above.
-  //           Notice that this can be done *before* submission.
-  // 
-  // TODO: Before this can be merged, the tx.claimableBalanceId(int) helper
-  //       actually needs to be implemented lol
-  .then(function(pair) {
-    // why do people put up with a language that doesn't have multiple
-    // return values in 2020 lmao
-    let tx = pair[0], txResponse = pair[1]
-    console.log("Claimable balance created!")
-
-    // let balanceId = tx.ClaimableBalanceID(0)
-    // console.log("Balance ID (1):", balanceId)
-    return txResponse
-  })
-  // Method 2: Suppose `txResponse` comes from the transaction submission
-  // above.
-  .then(function(txResponse) {
-    let txResult = sdk.xdr.TransactionResult.fromXDR(txResponse.result_xdr, "base64")
-    let results = txResult.result().results()
-
-    // We look at the first result since our first (and only) operation
-    // in the transaction was the CreateClaimableBalanceOp.
-    let operationResult = results[0].value().createClaimableBalanceResult()
-    let balanceId = operationResult.balanceId().toXDR("hex")
-
-    console.log("Balance ID (2):", balanceId)
-  })
-  // Method 3: Account B could alternatively do something like:
-  .then(async function() {
-    return await server
-      .claimableBalances()
-      .claimant(B.publicKey())
-      .call()
-      .then(function(balances) {
-        console.log(balances)
-        let balanceId = balances.records[0].id
-        console.log("Balance ID (3):", balanceId)
-        return balanceId
-      })
-      .catch(showAndThrow)
-  })
-  // Now Account B can actually claim the balance, provided too much time
-  // hasn't elapsed, or Account A can "reclaim" it otherwise.
-  .then(function(balanceId) {
-    console.log("Claiming", balanceId, "as", A.publicKey())
-    let claimBalance = sdk.Operation.claimClaimableBalance({
-      balanceId: balanceId
-    })
-
-    let tx = new sdk.TransactionBuilder(aAccount, // or Account B!
-      {
-        networkPassphrase: sdk.Networks.TESTNET,
-        fee: sdk.BASE_FEE,
-      })
-      .addOperation(claimBalance)
-      .setTimeout(180)
-      .build()
-
-    tx.sign(A)
-    return server.submitTransaction(tx)
-  })
-  .catch(check)
 ```
 
 ```go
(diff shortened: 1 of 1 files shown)


Shaptic:
The relevant page in the live preview is [here](https://developers-pr290.previews.kube001.services.stellar-ops.com/docs/glossary/claimable-balance/#example). 

Please sanity-check the JavaScript, @stellar/horizon-committers and @marcelosalloum; I'd like it to be as idiomatic as possible! Thanks in advance. :)

content/docs/glossary/claimable-balance.mdx
```diff
@@ -49,12 +49,57 @@ This operation will load the ClaimableBalanceEntry that corresponds to the Balan
...
+const sdk = require('stellar-sdk');
+
+var server = new sdk.Server("https://horizon-testnet.stellar.org")
+
+var A = sdk.Keypair.fromSecret("SAQLZCQA6AYUXK6JSKVPJ2MZ5K5IIABJOEQIG4RVBHX4PG2KMRKWXCHJ")
+var B = sdk.Keypair.fromPublicKey("GAS4V4O2B7DW5T7IQRPEEVCRXMDZESKISR7DVIGKZQYYV3OSQ5SH5LVP")
+
+// NOTE: Error checks are omitted for brevity; always validate things!
+
+server
```
tamirms, line 70: I think we prefer to use `await` instead of chaining promises

marcelosalloum: Personally I prefer Promises over Async/Await but we should stick to whatever is the most commonly used pattern in our official documentations.

marcelosalloum (commented):
LGTM! After we standardize the async function calls to use only one of `await` or `promises` we should be good to go.

content/docs/glossary/claimable-balance.mdx
```diff
@@ -49,12 +49,57 @@ This operation will load the ClaimableBalanceEntry that corresponds to the Balan
...
+    })
+
+    let tx = new sdk.TransactionBuilder(aAccount, {fee: sdk.BASE_FEE})
+      .addOperation(claimableBalanceEntry)
+      .setNetworkPassphrase(sdk.Networks.TESTNET)
+      .setTimeout(180)
+      .build()
+
+    tx.sign(A)
+    let txResponse = await server.submitTransaction(tx)
```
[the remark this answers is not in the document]
Shaptic: Clearly a knowledge gap on my part, thanks!

In general, how should one go about doing things that need waiting-on (like `server.loadAccount`, for example) in a `.then()` body? Or are those supposed to be delegated to their own `.then()` link in the Promise chain? In other words, would it have been more idiomatic of me to do something like 
```js
  return server.submitTransaction(tx)
}).then(function(txResponse) {
  console.log("Claimable balance created!")
  // The rest...
})
```
here, instead?

content/docs/glossary/claimable-balance.mdx
Shaptic: Tangentially-related to this: what's the best way to pass along multiple values from Promise to Promise? I was doing something like:
```js
  return [a, b]
}).then(function(pair) {
  let a = pair[0], b = pair[1]
  // the rest...
})
```
but that felt *really* odd to me. Are there better ways?

The reason I ask, for additional context, is because sometimes you want to use long-term values (like accounts) in addition to shorter-term values (like a transaction) in a single Promise body. Or is this already a flawed way of thinking?

Shaptic:
I believe I've resolved the mixed paradigm issue in 5a9a4a88d8a9937cc8a02a224facf5fe2c85f034, applying async/await across the board and removing all of the `.then()` chains! Please have another glance. :)

content/docs/glossary/claimable-balance.mdx
```diff
@@ -134,6 +179,58 @@ Either party could also check the `/effects` of the transaction, query `/claimab
...
+    let txResult = sdk.xdr.TransactionResult.fromXDR(txResponse.result_xdr, "base64")
+    let results = txResult.result().results()
+
+    // We look at the first result since our first (and only) operation
+    // in the transaction was the CreateClaimableBalanceOp.
+    let operationResult = results[0].value().createClaimableBalanceResult()
+    let balanceId = operationResult.balanceId().toXDR("hex")
+
+    console.log("Balance ID (2):", balanceId)
+  })
```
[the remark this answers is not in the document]
Shaptic: Added an extra blank line between each `// Method n: ` block!

content/docs/glossary/claimable-balance.mdx
marcelosalloum: In typescript you can return a typed object, but not so sure for javascript, it's been a few years since I did some heavy JS work.

content/docs/glossary/claimable-balance.mdx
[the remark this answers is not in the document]
marcelosalloum: If you go with promises you will want to maintain the depth of your code to 1 to avoid [hadouken code](https://www.reddit.com/r/ProgrammerHumor/comments/27yykv/indent_hadouken/).

**Desired promise structure**:
```js
server.submitTx(tx)     // async function
.then((result1) => {
  // do some stuff with result1 here if needed
  return server.applyResult1(result1)
})
.then((finalResult) => {
  // do some stuff with finalResult here if needed
  return server.finalize(finalResult)
})
.catch((err) => {
  handleErr(err)
})
```
*Please note the code above has depth=1. The depth can remain 1 regardless of the number of chained promises.*

**Avoidable hadouken code**:
```js
server.submitTx(tx)     // async function
.then((result1) => {
  // do some stuff with result1 here if needed
  server.applyResult1(result1)
  .then((finalResult) => {
    // do some stuff with finalResult here if needed
    return server.finalize(finalResult)
  })
})
.catch((err) => {
  handleErr(err)
})
```
*Note the code above has depth=2. The depth will increase proportionally with the number of chained promises.*

--
Of it doesn't make such a difference if the depth is 2 but it could escalate to something undesirable.

marcelosalloum (commented):
Looking good! You're just missing an important detail in the async tasks.

Usually async functions can throw errors so you need to apply try/catch around the `await` calls, like in https://javascript.info/async-await#error-handling. 

This was previously being handled by the promise `.catch()` method but you're not using promises anymore.

Also, I'm not so sure all async methods you're using will need try/catch but I'd guess they do 😅 . Better safe than sorry.

Shaptic:
Given the disclaimer:
> `// NOTE: Error checks are omitted for brevity; always validate things! `
What do you think about the "handling" in 558f807e? I threw it in for completeness sake, since I don't want to clutter the main point of the example, which is to demonstrate the actual SDK features. But I'm unsure about this, since maybe we still do want to demonstrate best practices regardless of example brevity.

Actually this led me to another thought, which would be to distinguish between the web-facing examples (minimal, segmented, etc.) and "complete" examples (compile+run-able, w/ error checks, abstractions, etc.). It'd be cool if we could create the latter and link it from the former. This would be much like how many tutorials have a "clone this project" button, and would let us separate "here's how you can use the feature" and "here's a proper implementation," since the latter is obviously far more involved and verbose.

marcelosalloum:
>Given the disclaimer:
>
>>`// NOTE: Error checks are omitted for brevity; always validate things!`

Sorry I missed that statement. I think you should either revert your last commit or update the code according with the following suggestions marked with a 💡 .

--

💡 FYI, you should encapsulate only the throwable code in `try/catch`. It's preferable to do:
```js
try {
  let result = await server.DoStuff()
} catch (err) {
  handleErr(err)
}
processResult(result)
doOtherStuff()
```

rather than:
```js
try {
  let result = await server.DoStuff()
  processResult(result)
  doOtherStuff()
} catch (err) {
  handleErr(err)
}
```
so an exception will be easier to debug.

--

💡 Also, when you use `try` without `catch` you're basically ignoring the error, which is more dangerous than not handling it. `try` and `catch` should always be used together.

Shaptic:
Awesome, I've added `await`-specific error-handling based on the `await promise().catch()` method described in [this section](https://medium.com/@vcarl/handling-errors-with-async-await-and-promises-cd2fea534f08#b1e5) in the latest push (also split out the latter part of the example into two sections for clarity, but that's purely aesthetic).

-----------

Separate question: what's the best way to handle the "success vs. failure" case when at the end of a function?
```js
return await server.submitTransaction(tx).then(function() {
  console.log("Claimable balance created!");
}).catch(function (err) {
  console.error(`Tx submission failed: ${err}`)
})
```

vs

```js
let txResponse = await server.submitTransaction(tx).catch(function (err) {
  console.error(`Tx submission failed: ${err}`)
})
if (!txResponse) return
console.log("Claimable balance created!");
return txResponse
```

The former feels cleaner (and is what I did in the changes), but I'm worried it dives back into "mixing `await` and `.then()`" territory. But then (heh) I wonder why mixing `await` and `.catch()` as described in the article is okay? Unless I'm conflating things again. Anyway, sorry for the mess of questions, but I think I'm getting a better grasp on JS and subsequent PRs should move forward faster.

content/docs/glossary/claimable-balance.mdx
```diff
@@ -49,12 +49,62 @@ This operation will load the ClaimableBalanceEntry that corresponds to the Balan
...
+    .setNetworkPassphrase(sdk.Networks.TESTNET)
+    .setTimeout(180)
+    .build();
+
+  tx.sign(A);
+  return await server.submitTransaction(tx).then(function() {
+    console.log("Claimable balance created!");
+  }).catch(function (err) {
+    console.error(`Tx submission failed: ${err}`)
+  });
```
marcelosalloum, line 104: You shouldn't return a value here since this example is running inside a `main()` function.

```suggestion
  await server.submitTransaction(tx).then(function() {
    console.log("Claimable balance created!");
  }).catch(function (err) {
    console.error(`Tx submission failed: ${err}`)
  });
```

content/docs/glossary/claimable-balance.mdx
```diff
@@ -158,9 +239,31 @@ if results, ok := txResult.OperationResults(); ok {
...
+tx = new sdk.TransactionBuilder(aAccount, {fee: sdk.BASE_FEE})
+  .addOperation(claimBalance)
+  .setNetworkPassphrase(sdk.Networks.TESTNET)
+  .setTimeout(180)
+  .build();
+
+tx.sign(A);
+return await server.submitTransaction(tx).catch(function (err) {
+  console.error(`Tx submission failed: ${err}`)
+});
```
marcelosalloum, line 263: This code is not inside any specific function in this example so you shouldn't be returning anything. An `await` without `return` is the way to go:

```suggestion
await server.submitTransaction(tx).catch(function (err) {
  console.error(`Tx submission failed: ${err}`)
});
```

marcelosalloum (commented):
There are a couple of unneeded returns ⬇️

content/docs/glossary/claimable-balance.mdx
Shaptic: Yep, excellent point. I think this is a copy-paste remnant from when I was actually *running* this code, before splitting it up into this "tutorial format".

marcelosalloum:
> Separate question: what's the best way to handle the "success vs. failure" case when at the end of a function?

You probably don't want to catch an error and return a _false success_ to the caller. Imagine a caller function is a view that calls `async changePassword(myCredentials)`, if there is an error inside the `changePassword` method the caller needs to know about it and show the appropriate message to the user. Promises and async functions are by definition throwable so you should (almost) never do something like this since it basically silences all errors:
```js
 return await server.submitTransaction(tx).then(function() {
   console.log("Claimable balance created!");
 }).catch(function (err) {
   console.error(`Tx submission failed: ${err}`)
 }) // uh oh, now the caller will think it was a success even when your method fails
```

---

About catching the error or not, that really depends on where you're calling the function from. 

If you're writing code in an SDK, i.e. in a **callee** that's been called by a **caller**, you'll want to propagate the error to the caller without handling it there, and it's the caller's responsibility to handle it:
```js
 return await server.submitTransaction(tx).then(function() {
   console.log("Claimable balance created!");
 }) // we're not catching the error here as we don't know what the caller might wanna do for each error type. Some times people want to catch and rethrow a different error, but that's an advanced topic for now.
```

If you are running the function from `main()` or from a view, it usually means you're the ultimate **caller** and you'll want to handle the error yourself:
```js
 await server.submitTransaction(tx).then(function() {
   console.log("Claimable balance created!");
 }).catch(function (err) {
   console.error(`Tx submission failed: ${err}`)
 }) // since you're the ultimate caller you won't return anything.
```

Just keep in mind someone (some method) needs to handle the error by either propagating to the caller, catching it and showing something to the user or catching it and moving to plan B.

marcelosalloum (approved):
LGTM!

Shaptic:
Got it! Makes a ton of sense. I keep forgetting about the "throwable" aspect (too much Go code lately...), so to me this model:
```js
 return await server.submitTransaction(tx).then(function() {
   console.log("Claimable balance created!");
 }).catch(function (err) {
   console.error(`Tx submission failed: ${err}`)
 }) // uh oh, now the caller will think it was a success even when your method fails
```
did not feel like a "false success" since the caller should do a null check :sweat_smile: . But obviously that's much easier to forget to do relative to an exception blowing up in your face. Great points all around, will keep them in mind! :+1:

rice2000:
Nice work!

PR #299, opened 2020-12-17
Update send-and-receive-payments.mdx

braduf:
Use txnbuild.MinBaseFee instead of MinBaseFee


rice2000 (approved):
Good eye!  Thanks for your help, @braduf.

PR #301, opened 2020-12-18
API: fix heading anchor

quietbits:
The issue was that `BasicLink` is not checking for everything we need for proper linking, that logic lives in `Link`.


quietbits:
This PR was breaking.

PR #302, opened 2020-12-18
Fail CI if package.json is updated but not yarn.lock

vcarl:


quietbits (approved):

Godspowerogun (commented):
For the confirmation of PowerPoint

PR #303, opened 2020-12-18
API: fix heading anchor

quietbits:
The issue was that `BasicLink` is not checking for everything we need for proper linking, that logic lives in `Link`.
Also, reverted `babel-plugin-styled-components` and `gatsby-plugin-styled-components` because they were breaking builds.


leighmcculloch (approved):

acharb (approved):
LGTM

PR #310, opened 2021-01-07
Remove exchange tutorial

tomerweller:
This exchange tutorial uses very confusing terminology ("issuing account"???) and I doubt any reputable exchange will find it useful. I vote to remove it.


rice2000 (approved):
Works for me.  If it's necessary, we can replace with an improved doc later.  But let's see.

PR #314, opened 2021-01-13
add sponsor to data entry

tomerweller:


tomerweller:
closing because I accidentally added this to the request instead of the response. where is the shape of the data entry response defined? only in the example? it feels weird adding an optional field to the general example

PR #328, opened 2021-02-04
Introduces a guide on error handling strategies.

Shaptic:
This should provide a comprehensive overview on the types of errors, certain common specific errors & how they can be resolved, and the general mindset necessary when working within the Stellar tech stack (SDKs, Horizon, Stellar-Core).

Tagging lots of people (basically anyone who was even briefly involved in the discussions) for review because we need diversity of perspective from different parts of the stack. Especially useful will be JavaScript experts and basically anyone who has had to deal with errors: @stellar/horizon-committers, @stellar/js-sdk-maintainer, @tyvdh, @JakeUrban, @tomerweller.

**The live preview can be viewed directly [here](https://developers-pr328.previews.kube001.services.stellar-ops.com/docs/tutorials/handling-errors/)** (might go down during pushes while it rebuilds).

### DIFF
--- a/content/docs/tutorials/handling-errors.mdx
+++ b/content/docs/tutorials/handling-errors.mdx
@@ -31,20 +31,21 @@ Just like with query errors, there are dozens of possible causes of a transactio
 
 Here are some of the common errors encountered when performing a `POST /transactions`:
 
-  - [400 error, `tx_bad_seq`](#invalid-sequences)
+  - [400 errors, e.g. `tx_bad_seq`](#invalid-sequences)
+  - [Fee / surge pricing errors, e.g. `tx_insufficient_fee`](#surge-pricing)
   - [504 error, timeout](#timeouts)
-  - [Surge pricing, `tx_insufficient_fee`](#surge-pricing)
   - [No HTTP response](#outages)
 
 Keep in mind this general rule of thumb: **any response other than `200 OK` or `400 Bad Request` means that the final status of the transaction is unknown.**
 
+
 ### 400 Errors
 Generally-speaking, if you get a 400 error during transaction submission, there are two main causes: either your transaction is in an unacceptable format or Stellar Core rejected your transaction. The former case is described by the [`transaction_malformed`](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-malformed/) error code and the latter by the [`transaction_failed`](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-failed/) code.
 
-Some transaction failures are recoverable; you should look at the `extras.result_code` field in the response message to determine the [root cause](https://developers.stellar.org/api/errors/result-codes/transactions/). One of these recoverable errors is `tx_bad_seq`, where the sequence number on the transaction doesn't match what's expected for the account.
+Some transaction failures are recoverable; you should look at the `extras.result_code` field in the response message to determine the [root cause](https://developers.stellar.org/api/errors/result-codes/transactions/). One of these recoverable errors is `tx_bad_seq`, where the sequence number on the transaction doesn't match what's expected for the account; another is `tx_insufficient_fee` which is covered [later](#surge-pricing).
 
 
-### Invalid Sequences
+#### Invalid Sequences
 These errors typically occur when you have an outdated view of an account due to other transactions happening outside of your worldview. This could be because the account is used on multiple devices, you have concurrent submissions happening, or a number of other reasons.
 
 Thankfully, the solution is relatively simple: retrieve the account details and try again with an updated sequence number.
@@ -72,28 +73,6 @@ server.submitTransaction(tx).catch(function (error)) {
 
 Despite the simple solution, something to be extremely careful about is **duplicate submissions**. Suppose you submitting transactions from multiple places in your application concurrently, and your user spammed the "Send Payment" button a few times in impatience. If you try to send the *exact same* payment transaction for each tap, naturally only one of them will succeed. However, since the others fail with `tx_bad_seq`, the above code will blindly resubmit them with an updated sequence number and **also succeed**, ultimately resulting in more than one payment being made when only one was intended. In summary, be careful with resubmissions and recoverable transaction errors.
 
-
-
-
-### Timeouts
-If you receive a 504 error from Horizon after a transaction submission, you've encountered a timeout. In the Stellar universe, a timeout indicates something a little nebulous and non-traditional: it's more of a *warning* that your request hasn't been fulfilled within a reasonable amount of time *yet* more than an error. This subtlety arises because of the nature of the relationship between Horizon and Stellar Core: the network might take some time to accept a transaction---especially one with a low fee, see [Surge Pricing](#surge-pricing) later---on the order of 5-10 minutes in the worst cases, whereas Horizon needs to provide developers with a response within 30 seconds or so.
-
-This leads to a very important point: **receiving a 504 for your transaction submission does not mean the transaction did not make it to the network**.
-
-There's no mechanism to "cancel" a transaction after it has been submitted aside from by exceeding its timebounds. In fact, one of the best tools available for managing transaction failures is through [timebounds](https://developers.stellar.org/docs/glossary/transactions/#time-bounds). While optional, they're *crucial* in simplifying error handling: it's much easier to know when to check a transaction's status when you know the maximum amount of time it can be stuck.
-
-With this trick in our back pocket, resolving 504s is straightforward: until the timebound has expired, you can safely retry submission *without modifying the transaction*. Horizon makes an important guarantee (excerpt is from [this example](https://developers.stellar.org/docs/tutorials/send-and-receive-payments/#send-a-payment)) for this scenario:
-
-> If the transaction has already been successfully applied to the ledger, Horizon will simply return the saved result and not attempt to submit the transaction again. Only in cases where a transaction’s status is unknown (and thus will have a chance of being included into a ledger) will a resubmission to the network occur.
-
-In other words, if your transaction made it through after the 504, you'll get your previously-submitted transaction back. If it didn't (e.g. a `tx_too_late` result code), you're essentially back to square one: you might 504 again, fail for another reason (e.g. `tx_bad_seq`), or succeed.
-
-#### Transaction Bans
-Note the following caveat of the [transaction lifecycle](https://developers.stellar.org/docs/glossary/transactions/#transaction-lifecycle): 
-
-> If a transaction is kept in memory after a certain number of ledger closes, **it will be banned for several additional ledgers.** This means no attempt will be made to include it in a candidate transaction set additional ledgers during this time.
-
-If you trigger a ban scenario while your timebound is still in the future, you may have no choice but to use a [fee-bump transaction](#bumping-fees-on-past-transactions) to get into the ledger.
 
 #### Surge Pricing
 When the Stellar network undergoes bursts of activity, [surge pricing](https://developers.stellar.org/docs/glossary/fees/#surge-pricing) might kick in. This and other fee fluctuations can cause unexpected errors during transaction submission in applications that don't plan ahead for its dynamic nature.
@@ -174,6 +153,27 @@ server.submitTransaction(lastTx).catch(function (error) {
 Note an [important stipulation](https://developers.stellar.org/docs/glossary/fee-bumps/#replace-by-fee) of fee bumping that's fulfilled above:
 
 > If you submit two distinct transactions with the same source account and sequence number, and the second transaction is a fee-bump transaction, the second transaction will be included in the transaction queue in place of the first transaction if and only if **the fee bid of the second transaction is at least 10x the fee bid of the first transaction**.
+
+
+### Timeouts
+If you receive a 504 error from Horizon after a transaction submission, you've encountered a timeout. In the Stellar universe, a timeout indicates something a little nebulous and non-traditional: it's more of a *warning* that your request hasn't been fulfilled within a reasonable amount of time *yet* more than an error. This subtlety arises because of the nature of the relationship between Horizon and Stellar Core: the network might take some time to accept a transaction---especially one with a low fee, see [Surge Pricing](#surge-pricing) later---on the order of 5-10 minutes in the worst cases, whereas Horizon needs to provide developers with a response within 30 seconds or so.
+
+This leads to a very important point: **receiving a 504 for your transaction submission does not mean the transaction did not make it to the network**.
+
+There's no mechanism to "cancel" a transaction after it has been submitted aside from by exceeding its timebounds. In fact, one of the best tools available for managing transaction failures is through [timebounds](https://developers.stellar.org/docs/glossary/transactions/#time-bounds). While optional, they're *crucial* in simplifying error handling: it's much easier to know when to check a transaction's status when you know the maximum amount of time it can be stuck.
+
+With this trick in our back pocket, resolving 504s is straightforward: until the timebound has expired, you can safely retry submission *without modifying the transaction*. Horizon makes an important guarantee (excerpt is from [this example](https://developers.stellar.org/docs/tutorials/send-and-receive-payments/#send-a-payment)) for this scenario:
+
+> If the transaction has already been successfully applied to the ledger, Horizon will simply return the saved result and not attempt to submit the transaction again. Only in cases where a transaction’s status is unknown (and thus will have a chance of being included into a ledger) will a resubmission to the network occur.
+
+In other words, if your transaction made it through after the 504, you'll get your previously-submitted transaction back. If it didn't (e.g. a `tx_too_late` result code), you're essentially back to square one: you might 504 again, fail for another reason (e.g. `tx_bad_seq`), or succeed.
+
+#### Transaction Bans
+Note the following caveat of the [transaction lifecycle](https://developers.stellar.org/docs/glossary/transactions/#transaction-lifecycle): 
+
+> If a transaction is kept in memory after a certain number of ledger closes, **it will be banned for several additional ledgers.** This means no attempt will be made to include it in a candidate transaction set additional ledgers during this time.
+
+If you trigger a ban scenario while your timebound is still in the future, you may have no choice but to use a [fee-bump transaction](#bumping-fees-on-past-transactions) to get into the ledger.
 
 
 ### Outages
--- a/content/docs/tutorials/handling-errors.mdx
+++ b/content/docs/tutorials/handling-errors.mdx
@@ -142,7 +142,7 @@ Despite the solution's simplicity, things can go very wrong very fast if you don
 
 Suppose you submit transactions from multiple places in your application concurrently, and your user spammed the "Send Payment" button a few times in their impatience. If you try to send the *exact same* payment transaction for each tap, naturally only one will succeed. The others will fail with an invalid sequence number (`tx_bad_seq`), and if you resubmit blindly with an updated sequence number (as we do above), these payments will **also succeed**, ultimately resulting in more than one payment being made when only one was intended.
 
-In essence, be very careful with resubmitting transactions that have been modified to work around an error.
+In essence, **be very careful with resubmitting transactions that have been modified** to work around an error.
 
 
 
@@ -151,7 +151,7 @@ This section covers a smattering of specific error cases commonly encountered du
 
 
 ### Timeouts
-If you receive a `504 Timeout` from Horizon after a transaction submission, you've encountered something a little nebulous and non-traditional in the Stellar universe: timeouts are more of a *warning* that your request hasn't been fulfilled within a reasonable amount of time *yet* rather than an error. This subtlety arises because of the nature of the relationship between Horizon and Stellar Core: the network might take some time to accept a transaction---especially one with a low fee, see [Surge Pricing](#insufficient-fees-and-surge-pricing) later---on the order of 5-10 minutes during congestion, whereas Horizon needs to provide developers with a response within a reasonable 30 seconds or so.
+If you receive a `504 Timeout` from Horizon after a transaction submission, you've encountered something a little nebulous and non-traditional in the Stellar universe: timeouts are more of a *warning* that your request hasn't been fulfilled within a reasonable amount of time *yet* rather than an error. This subtlety arises because of the nature of the relationship between Horizon and Stellar Core: the network might take some time to accept a transaction&mdash;especially one with a low fee, see the discussion of [Surge Pricing](#insufficient-fees-and-surge-pricing) later&mdash;on the order of 5-10 minutes during congestion, whereas Horizon needs to provide developers with a response within a reasonable 30 seconds or so.
 
 This leads to a very important point: **receiving a 504 for your transaction submission does not mean the transaction did not make it to the network**. You should continue with [retries](#retrying-until-success) until getting a more definitive response.
 
--- a/content/docs/tutorials/handling-errors.mdx
+++ b/content/docs/tutorials/handling-errors.mdx
@@ -1,5 +1,5 @@
 ---
-title: Securing Web-based Projects
+title: Handling Errors Gracefully
 order: 100
 ---
 
--- a/content/docs/tutorials/handling-errors.mdx
+++ b/content/docs/tutorials/handling-errors.mdx
@@ -5,7 +5,7 @@ order: 100
 
 import { CodeExample } from "components/CodeExample";
 
-When working with a payment network like Stellar, expecting the unexpected is critical in ensuring a good user experience: one in which money doesn't get lost, actions happen when they're supposed to (or as early as they can), and everyone's view of the network is as accurate as possible.
+When working with a payment network like Stellar[,](https://www.youtube.com/watch/TLXo9IFVboY) expecting the unexpected is critical in ensuring a good user experience: one in which money doesn't get lost, actions happen when they're supposed to (or as early as they can), and everyone's view of the network is as accurate as possible.
 
 In many of the tutorials and code samples throughout this documentation, we've minimized the error handling code to reduce verbosity and focus on the essence of the examples. Here, we'll do the exact opposite, because error handling *is* the essence: by the end, you should be able to categorize errors and understand the best, most-idiomatic way to handle them in your application.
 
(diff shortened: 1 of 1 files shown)


content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,272 @@
...
+}
+```
+
+</CodeExample>
+
+
+## Invalid Sequences
+These errors typically occur when you have an outdated view of an account due to other transactions happening outside of your worldview. This could be because the account is used on multiple devices, you have concurrent submissions happening, or a number of other reasons.
+
+Thankfully, the solution is relatively simple: retrieve the account details and try again with an updated sequence number.
```
bartekn, line 112: I think it's worth noting here that updating sequence number to the latest value can sometime cause a double spend. Say you have two submitters running and they both try to send the same payment tx with the same seqnum. One of them will succeed, the other will get bad sequence error and will try to submit exactly the same tx again. If it succeeds this time the payment will be sent twice.

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+```js
+// Let `lastTx` be some transaction that fails submission due to high fees, and
+// `lastFee` be the maximum fee (expressed as a float) willing to be paid by 
+// `account` for `lastTx`.
+server.submitTransaction(lastTx).catch(function (error) {
+  if (isFeeError(error)) {
+    let bump = sdk.TransactionBuilder.buildFeeBumpTransaction(
+      account, (lastFee * 10).toFixed(2), lastTx, server.networkPassphrase
+    );
```
bartekn, line 198: This isn't clear to me. The fee should be an integer (in stroops) but we use `toFixed` here. Also adding a suggestion that will be more readable:
```suggestion
    let bump = sdk.TransactionBuilder.buildFeeBumpTransaction(
      account, // Account that will pay the fee
      (lastFee * 10).toFixed(2), // New fee
       lastTx,
       server.networkPassphrase
    );
```

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,272 @@
...
+It's worth noting that SDKs commonly classify 404s uniquely relative to other error codes (e.g. `NotFoundError` vs the generic `NetworkError` in the JavaScript SDK).
+
+## Resolutions
+TODO: Either fix the parameters (in case of user error, insert link to API reference) or try again (in case of transient error). 
+
+Are there other cases?
+
+
+
+# Transaction Submission Errors
```
bartekn, line 75: I think it's worth adding a general rule of thumb here: any response other than `200 OK` or `400 Bad Request` means that the final status of the transaction is unknown.

Also, would be great to create a section with a generic error handling algorithm in pseudocode.

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+With this trick in our back pocket, resolving 504s is straightforward: until the timebound has expired, you can safely retry submission *without modifying the transaction*. Horizon makes an important guarantee (excerpt is from [this example](https://developers.stellar.org/docs/tutorials/send-and-receive-payments/#send-a-payment)) for this scenario:
+
+> If the transaction has already been successfully applied to the ledger, Horizon will simply return the saved result and not attempt to submit the transaction again. Only in cases where a transaction’s status is unknown (and thus will have a chance of being included into a ledger) will a resubmission to the network occur.
+
+In other words, if your transaction made it through after the 504, you'll get your previously-submitted transaction back. If it didn't, you've essentially started from scratch again: you might 504 again, fail for another reason (e.g. `tx_bad_seq`), or succeed.
```
bartekn, line 226:
```suggestion
In other words, if your transaction made it through after the 504, you'll get your previously-submitted transaction's result back. If it didn't, you've essentially started from scratch again: you might 504 again, fail for another reason (e.g. `tx_bad_seq`), or succeed.
```

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+
+## Outages
+Receiving no response from Horizon after a submission is very similar to receiving a 504: you're uncertain about the outcome of the transaction submission.
+
+It's obviously crucial to determine whether or not it went through or not, but you can likewise just keep trying submission until the timebounds expire. Once they have, though, if the transient error hasn't resolved itself, you should just stop and wait; submitting a new transaction might result in a undesirable duplicate action.
```
bartekn, line 232: > just keep trying submission until the timebounds expire

I'm not sure if we want to complicate it more but _keep trying_ no longer works with banning in stellar-core. So after a few ledgers the only option may be to send fee bump tx.

Shaptic: You need to update timebounds even w/ a fee bump, right? So the txhash would differ. So you may cause a double-spend? Just making sure this is accurate so I note it.

content/docs/tutorials/handling-errors.mdx
Shaptic: Good point!

content/docs/tutorials/handling-errors.mdx
Shaptic: Good point! It's supposed to be an integer but encoded as a string; I'll rectify this! `.toFixed` is necessary to avoid long decimals (e.g. `1000.000000001`) but you're right that there shouldn't be a decimal in the first place.

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+import { CodeExample } from "components/CodeExample";
+
+When working with a payment network like Stellar, expecting the unexpected is critical in ensuring a good user experience: one in which money doesn't get lost, actions happen when they're supposed to (or as early as they can), and everyone's view of the network is as accurate as possible.
+
+In many of the tutorials and code samples throughout this documentation, we've minimized the error handling code to reduce verbosity and focus on the essence of the examples. Here, we'll do the exact opposite, because error handling *is* the essence: by the end, you should be able to categorize errors and understand the best, most-idiomatic way to handle them in your application.
+
+Since [the SDKs](../software-and-sdks/) are the main way people interact with the Stellar network, their error reporting will be the main focus of this document. At the end of the day, though, they're just human- and API-friendly abstractions on the underlying problems. As such, the knowledge should be relatively easy to transfer to other scenarios such as using a 3rd party SDK or interfacing with Horizon directly over HTTP.
+
+# Contents
```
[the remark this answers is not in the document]
Shaptic: Good point; I didn't even realize there was an outline, which removes the need for the ToC.

content/docs/tutorials/handling-errors.mdx
bartekn: I'm talking about the case in which `timebounds.max_time` is still in the future. When it's in the past and tx is not in the ledger you can create a new tx with the same seqnum.

content/docs/tutorials/handling-errors.mdx
Shaptic: > Also, would be great to create a section with a generic error handling algorithm in pseudocode.

TBD in a future PR!

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+With this trick in our back pocket, resolving 504s is straightforward: until the timebound has expired, you can safely retry submission *without modifying the transaction*. Horizon makes an important guarantee (excerpt is from [this example](https://developers.stellar.org/docs/tutorials/send-and-receive-payments/#send-a-payment)) for this scenario:
+
+> If the transaction has already been successfully applied to the ledger, Horizon will simply return the saved result and not attempt to submit the transaction again. Only in cases where a transaction’s status is unknown (and thus will have a chance of being included into a ledger) will a resubmission to the network occur.
+
+In other words, if your transaction made it through after the 504, you'll get your previously-submitted transaction back. If it didn't (e.g. a `tx_too_late` result code), you're essentially back to square one: you might 504 again, fail for another reason (e.g. `tx_bad_seq`), or succeed.
```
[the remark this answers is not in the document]
JakeUrban: I didn't realize `504` meant core was still deciding -- I figured it had been rejected and that was that.

This brings me to another question. Lets say I get a 504. This means my fee is above the minimum so I didn't get a `tx_insufficient_fee`, but its not competitive with the other tx's being submitted, right? Is there a non-fee-related issue that would result in getting a `504` back?

If not, then we should state that `504`s mean the transaction's was too low relative to the other transactions being submitted, and maybe explain how that differs from `tx_insufficient_fee`. 

If there are other scenarios, is there a way to figure out what the reason was, if it wasn't related to fees?

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+#### Transaction Bans
+Note the following caveat of the [transaction lifecycle](https://developers.stellar.org/docs/glossary/transactions/#transaction-lifecycle): 
+
+> If a transaction is kept in memory after a certain number of ledger closes, **it will be banned for several additional ledgers.** This means no attempt will be made to include it in a candidate transaction set additional ledgers during this time.
```
JakeUrban, line 174: How does one end up causing this situation? If I get a `504`, and I keep resubmitting, am I risking the transaction getting banned?

JakeUrban (changes requested):
This is really coming together! I left some comments, mostly about being as direct/explicit as possible.

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,281 @@
...
+
+
+### Timeouts
+If you receive a 504 error from Horizon after a transaction submission, you've encountered a timeout. In the Stellar universe, a timeout indicates something a little nebulous and non-traditional: it's more of a *warning* that your request hasn't been fulfilled within a reasonable amount of time *yet* more than an error. This subtlety arises because of the nature of the relationship between Horizon and Stellar Core: the network might take some time to accept a transaction---especially one with a low fee, see [Surge Pricing](#surge-pricing) later---on the order of 5-10 minutes in the worst cases, whereas Horizon needs to provide developers with a response within 30 seconds or so.
+
+This leads to a very important point: **receiving a 504 for your transaction submission does not mean the transaction did not make it to the network**.
```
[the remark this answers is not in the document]
acharb: TIL x2! Great info here 💯 

question @Shaptic , if you don’t set timebounds on a transaction what’s the upperbound on the time Core will try to add your transaction to the ledger? (eg. if theoretically there are always competing transactions with higher baseFees during surge pricing, when does Core give up on your transaction)

I see you put `5-10 minutes` is there a hard max value for that? Either way I think adding a link or short note on Core's max effort for your transaction would be a helpful note here

content/docs/tutorials/handling-errors.mdx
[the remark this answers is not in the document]
Shaptic: > Is there a non-fee-related issue that would result in getting a 504 back?

Sure, there are plenty of transient issues that could cause it, I think. From my understanding, a `504 Timeout` **might** mean that Core is still deciding; it could also be due to a problem between Horizon and Core for example, or the Core node and the rest of the Stellar network. 

It's hard (if not impossible) to determine those error details, which is why tx resub until a hard success/failure is basically the only thing you can do.

content/docs/tutorials/handling-errors.mdx
Shaptic: From my understanding, bans happen if a tx stays in the queue for too long (others are taking priority due to higher fees). I've clarified this!

Shaptic:
Thank you for your (awesome) comments @JakeUrban; I'll be clarifying all of the points you brought up in the next push as soon as I optimize the new structure for the document to be "resolution-oriented" as suggested.

content/docs/tutorials/handling-errors.mdx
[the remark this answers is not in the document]
Shaptic: I think you got your answer in Slack @acharb, but for posterity, it appears that transactions w/o timebounds will **typically** be considered for 3 ledgers then banned for 10, but this can go on forever.

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,281 @@
...
+  - [No HTTP response](#outages)
+
+Keep in mind this general rule of thumb: **any response other than `200 OK` or `400 Bad Request` means that the final status of the transaction is unknown.**
+
+
+### 400 Errors
+Generally-speaking, if you get a 400 error during transaction submission, there are two main causes: either your transaction is in an unacceptable format or Stellar Core rejected your transaction. The former case is described by the [`transaction_malformed`](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-malformed/) error code and the latter by the [`transaction_failed`](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-failed/) code.
+
+Some transaction failures are recoverable; you should look at the `extras.result_code` field in the response message to determine the [root cause](https://developers.stellar.org/api/errors/result-codes/transactions/). One of these recoverable errors is `tx_bad_seq`, where the sequence number on the transaction doesn't match what's expected for the account; another is `tx_insufficient_fee` which is covered [later](#surge-pricing).
```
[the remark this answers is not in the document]
Shaptic: Good point, but I want to avoid enumerating errors if possible, since that should be *more-or-less* researchable from the [`result_codes` page itself](https://developers.stellar.org/api/errors/result-codes/transactions/). I don't think it's possible to be exhaustive. You're right, though; I'll think about this some more!

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,241 @@
...
+As outlined in the section on [retries](#retrying-until-success), resubmitting an *unchanged* (and valid) transaction (with the same operations, signatures, sequence number, etc.) is always safe to do. You should be careful when working around an error that *does* require changes to the transaction, though: it's very possible to cause **duplication transactions** which can result in all sorts of problems (double-payments, erroneous trustlines, etc.).
+
+##### Example: Invalid Sequence Numbers
+These errors typically occur when you have an outdated view of an account due to other transactions happening outside of your worldview. This could be because the account is used on multiple devices, you have concurrent submissions happening, or a number of other reasons. Thankfully, the solution is **usually** relatively simple: retrieve the account details and try again with an updated sequence number.
+
+<CodeExample>
+
+```js
+// suppose `account` is an outdated `AccountResponse` object
+let tx = sdk.TransactionBuilder(account, ...)/* etc */.build()
```
vcarl, line 122: The `...` here isn't valid JS, which might break some random syntax highlighting bits. Not the worst thing in the world, but thought I'd point it out while looking!

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,241 @@
...
+    return sleep(timeout).then(function () {
+      return submitTransaction(tx, timeout + 5);
+    });
+  });
+}
+```
+
+(We assume the existence of a `sleep` implementation akin to the one [here](https://stackoverflow.com/a/39914235).)
+
+</CodeExample>
```
vcarl, line 71: Can't have text inside a `CodeExample`, I think this is probably what's causing the error
```suggestion
</CodeExample>

(We assume the existence of a `sleep` implementation akin to the one [here](https://stackoverflow.com/a/39914235).)
```

Shaptic: Dunno how I overlooked this.. thank you!

content/docs/tutorials/handling-errors.mdx
Shaptic: Looks like it renders okay, but thanks for noting it! :+1:

MonsieurNicolas:
This looks good!

I think that in the "retrying until success" section, the information that you can also reach a "definite no" answer is a bit buried as well as the need for timeout.

I'd suggest extracting a new section "Preparing a *new* transaction for submission" (or better title), before "retrying...".

That way you can expand a bit on how to pick the expiration time value. For many "interactive apps" it's probably something between 30s-1m, for others it's OK to have expiration times much further in the future.
You can also add a note that if the application can crash/be forced closed (like mobile apps), code should be written in such a way that it stores the transaction locally (and uses that) before submitting it to the network as to avoid problems like the ones in the later section on duplicate transactions.

In the "Insufficient Fees and Surge Pricing" section, getting an error `TX_INSUFFICIENT_FEE` is not necessarily "final" depending on how people called APIs (if on a retry, getting that error may just mean that network conditions changed as on retry a different Horizon node may be called). Not for this doc, but I wonder if Horizon caches that error 🤔 
The only time you can be sure that a transaction was not included in a ledger is by observing a ledger with time passed the timebound of the transaction (same thing, I don't know if we can provide strong guarantees when calling Horizon as I don't think we have information on the current ledger in the return code of a transaction call - we may want to open issues to discuss this).

Related, in the "unsafe transaction adjustments", we have something similar: it's actually safe to change anything in a transaction *but* the transaction source account and sequence number pair. In particular, if there was some error like `TX_INSUFFICIENT_FEE` it's actually fine to just try to change the fee directly (without a fee bump - it's just not as effective to "unstick" transactions), or to modify signatures.
What can happen though is that *any* variations that share the same transaction source account + sequence number pair may "win", not just the last one. This complicates code in the client as now instead of "retrying until timeout or success" a single transaction, all of them have to be retried.

Shaptic:
Great points @MonsieurNicolas, and we'll be discussing the Horizon-related ones with the team.

I'll definitely add some text clarifying the nuances of sequence bumps vs. all other changes. Noting that the clients should track all `n` of their submissions since any of them have a chance of success (barring "true" error codes like `tx_bad_auth` or `tx_too_late`), regardless of fee adjustments (which only increase *likelihood* of success, afaict), is definitely important here.

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,243 @@
...
+  "reason": "Invalid claimable balance ID"
+}
+```
+
+</CodeExample>
+
+Note that the SDKs make it a point to distinguish an **invalid request** (as above) vs. a **missing resource** (a `404 Not Found`) (e.g. the generic `NetworkError` vs. a `NotFoundError` in the JavaScript SDK), where the latter might not be considered an error depending on your situation.
+
+#### Transaction Submissions
+Certain transaction submission failures also need adjustments to succeed. If the XDR is malformed or the transaction is somehow otherwise invalid, you'll encounter a `400 Bad Request` (for example, when excluding a source account as seen in the [API reference example](../../api/errors/response/)). Both transactions and their requisite operations can easily be malformed: ook at the `extras.result_codes` field for details and cross-reference them with the appropriate [Result Codes documentation](../../api/errors/result-codes/) to determine specifics.
```
rice2000, line 50:
```suggestion
Certain transaction submission failures also need adjustments to succeed. If the XDR is malformed or the transaction is somehow otherwise invalid, you'll encounter a `400 Bad Request` (for example, when excluding a source account as seen in the [API reference example](../../api/errors/response/)). Both transactions and their requisite operations can easily be malformed: look at the `extras.result_codes` field for details and cross-reference them with the appropriate [Result Codes documentation](../../api/errors/result-codes/) to determine specifics.
```

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,243 @@
...
+
+> If the transaction has already been successfully applied to the ledger, Horizon will simply return the saved result and not attempt to submit the transaction again. Only in cases where a transaction’s status is unknown (and thus will have a chance of being included into a ledger) will a resubmission to the network occur.  
+
+_(The above excerpt is from [this example](https://developers.stellar.org/docs/tutorials/send-and-receive-payments/#send-a-payment).)_
+
+In other words, if your transaction makes it through at any point in time, you'll get your previously-submitted transaction back for all subsequent retries.
```
rice2000, line 68:
```suggestion
In other words, if your transaction makes it through at any point in time, you'll get your previously submitted transaction back for all subsequent retries.
```

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,243 @@
...
+  });
+}
+```
+
+</CodeExample>
+
+_(We assume the existence of a `sleep` implementation akin to the one [here](https://stackoverflow.com/a/39914235).)_
+
+#### Details: Retry Backoff
+Be sure to integrate backoff into your retry mechanism. In our example error-handling code above, we implement a simple linear back-off, but there are [plenty of recommendations](https://backoff-utils.readthedocs.io/en/latest/strategies.html#why-are-backoff-strategies-useful) out there for various other strategies you can employ. Back-off is important both for maintaining performance and avoiding [rate-limiting](#rate-limiting) issues.
```
rice2000, line 105:
```suggestion
Be sure to integrate backoff into your retry mechanism. In our example error-handling code above, we implement a simple linear backoff, but there are [plenty of recommendations](https://backoff-utils.readthedocs.io/en/latest/strategies.html#why-are-backoff-strategies-useful) out there for various other strategies you can employ. Backoff is important both for maintaining performance and avoiding [rate-limiting](#rate-limiting) issues.
```

content/docs/tutorials/handling-errors.mdx
```diff
@@ -0,0 +1,243 @@
...
+
+  1. **Track fee fluctuations** via the [`fee_stats` endpoint](https://developers.stellar.org/api/aggregations/fee-stats/). This can let you make informed, specific choices about the fee you're comfortable paying. Alternatively, simply...
+
+  2. **Set the highest fee** you are comfortable with. Crucially, you should remember that *this doesn't mean you'll **pay** that on every transaction*. You will only pay whatever is necessary to get you into the ledger: under normal (non-surge) circumstances, even with a higher *maximum* fee set, you will pay the *standard* fee (100 stroops as of this writing).
+
+The latter strategy balances simplicity, efficacy, and convenience, but unless you set your maximum high enough simply *never* be exceeded, can still lead to failures. The former strategy can provide more-reliable submissions by allowing tighter guarantees about whether or not a transaction will be accepted. In general, though, it's important to track fee costs: if the network saturates beyond your maximum willingness to pay, perhaps waiting for activity to die down or periodically retrying with the same fee is the best approach for your use case.
```
rice2000, line 168:
```suggestion
The latter strategy balances simplicity, efficacy, and convenience, but unless you set your maximum high enough so that it's *never* exceeded, it can still lead to failures. The former strategy can provide more reliable submissions by allowing tighter guarantees about whether or not a transaction will be accepted. In general, though, it's important to track fee costs: if the network saturates beyond your maximum willingness to pay, perhaps waiting for activity to die down or periodically retrying with the same fee is the best approach for your use case.
```

rice2000 (commented):
Gave it a read and noted a few minor copyedits along the way.

content/docs/tutorials/handling-errors.mdx
JakeUrban: So as I understand, a transaction will stay in the queue until one of the following happens:
- it is included in the ledger (success)
- its upper timebound is exceeded (tx_too_late)
- its banned

Please correct me if I'm wrong but this raises the following questions:
- is there a fixed limit on how long a tx can stay in the queue (# of ledgers / # of seconds)? This limit should be mentioned if it exists.
- what is the Horizon response when I try to submit a banned transaction? I don't see a result code that directly applies
- Lets say every transaction submitted has a higher fee than the one I submit, and I don't include an upper timebound. The TX _will_ be banned right?

JakeUrban: Oh I see you answered my first question here: https://github.com/stellar/new-docs/pull/328#discussion_r579516471

But you say "typically". If I want to avoid getting banned (because I don't want to pay the 10x required to be reconsidered), what guidance would you give me? Always set an upper timebound of no more than 15 seconds (~3 ledgers)? I think we should provide some guidance here in the doc, because getting banned is a significant penalty.

Shaptic: I believe there's no limit on staying in the queue aside from what's achievable via the tx timebound. Here's what @MonsieurNicolas had to say on the matter:

> You need to set your timeouts based on your business requirements and assume that transactions may get dropped or re-processed at any time (transactions may sit around in queues for much longer than 3 ledgers). The only thing you can rely on are close time in the latest ledger that was closed and sequence number of your account in that latest ledger, then you need to design a sequence of transactions that deals with those races properly.

If you want to avoid a ban, the best strategy is (_likely_-I'm doing some inferring here) just to set your fees appropriately. For example, I can't imagine setting a max fee based on the 90th percentile (from `/fee_stats`) will ever lead to getting banned. I'll try to get answers for your other bullets from Core.

Shaptic: Despite our mutual curiosity and desire for clarity, @JakeUrban, I think it's actually best to avoid any mention of tx bans in this guide entirely (see fd87d6b68cb7f4ccd6748c58f57c05dd6d94e20d).

This thought process is magnified by the fact that a txban-specific error doesn't bubble up to SDK users and so represents an error scenario that's fairly deep and specific. As far as users are concerned, the resolution strategy stays the same: **set the max fee you're willing to pay** (or make a fee-bump / check `/fee_stats`) **and try again**. I think "exposing" this (deep) knowledge about Stellar Core's behavior (a) isn't actually helpful (except to satisfy curiosity) (b) strips away more of the (intended) abstraction of Horizon <--> Core and (c) locks Core into certain expectations. 

What do you think?

JakeUrban: I don't think we need to delay merging this doc over this conversation, but I do think it exposed some gaps in our understanding. 

It's not clear how Horizon handles Core's `TRY_AGAIN_LATER` error and how that is relayed to the client. It may be the case that Horizon lumps in into other error modes described in this doc (`504`), in which case the client truly doesn't need to worry about it. But I wanted to confirm that there isn't some undocumented error response that represents a tx ban that clients wouldn't be prepared to handle.

Shaptic:
> I'd suggest extracting a new section "Preparing a new transaction for submission" (or better title), before "retrying...".
>  
> That way you can expand a bit on how to pick the expiration time value. For many "interactive apps" it's probably something between 30s-1m, for others it's OK to have expiration times much further in the future.
You can also add a note that if the application can crash/be forced closed (like mobile apps), code should be written in such a way that it stores the transaction locally (and uses that) before submitting it to the network as to avoid problems like the ones in the later section on duplicate transactions.

@MonsieurNicolas btw I think this is a great suggestion, though I think it belongs in a different document given that this is specifically catered towards error handling. A "Transaction Submission Strategies" page made in a new PR that cross-references and back-links to this one is a better way to go, imo.

JakeUrban (approved):

PR #338, opened 2021-02-11
Fix JS example in follow-received-payments documentation

Shaptic:
The example needs to be updated to the latest SDK version. This was done in response to stellar/js-stellar-base#387.


rice2000 (approved):
Thank you!

PR #343, opened 2021-02-18
Clarify asset format for the claimable balance APIs

Shaptic:
This is a small change to unify the way the [CreateClaimableBalance operation](https://developers.stellar.org/api/resources/operations/object/create-claimable-balance/) is presented with the [object itself](https://developers.stellar.org/api/resources/claimablebalances/object/).


rice2000 (approved):
Nice!

PR #346, opened 2021-02-24
Clarify SDK expectations for Unix epochs in claimable balance predicates.

### LINKED ISSUE #345: Feature Request: Specify that claimable absolute time predicate must be seconds
### What problem does your feature solve?
Avoid misunderstanding between a time expressed in milliseconds and seconds.

### What would you like to see?
Would be a good idea to specify that the absolute time predicate (claimable balance op.) is expressed in seconds,
at the moment is indicated as "generic" Unix Epoch.

Shaptic:
Closes #345.


rice2000 (approved):
Nice!  Thank you.

PR #350, opened 2021-02-26
Remove references to non-existent file under /issuing-assets to fix 404s

shanzzam:
The file documenting how to list an asset on the DEX is currently under construction. As a result, links to the page  /issuing-assets/list-asset-on-dex.mdx results in 404s. Until this resource is created, remove links to this asset.


rice2000 (approved):
Looks good!  Thanks.

PR #351, opened 2021-03-03
Specify a string rather than a number for /assets?cursor={paging_token}

Shaptic:
Coupled with stellar/go#3440.


rice2000 (approved):
LGTM!

PR #353, opened 2021-03-05
Resolve misunderstanding: ingestion is NOT enabled by default in 2.0+

Shaptic:
Only ingestion **via Captive Core** is enabled by default, but users still need to choose to ingest.


bartekn (approved):

rice2000 (approved):
Good catch!

PR #358, opened 2021-03-12
Restore the *old* Horizon deployment method while we stabilize the 2.0+ docs.

Shaptic:
Per requests & recommendations on Slack, we're restoring the documentation that outlines the old way of deploying Horizon in order to ensure that users still have a (relatively) seamless way to set up Horizon if they encounter issues in the new, Horizon 2.x+, Captive Core-based documentation.

This section should not live for very long: we're aiming to maximize the user experience of the Captive Core deployment and drop this section again as soon as enough feedback comes in identifying hiccups and pain points.

### DIFF
--- a/content/docs/run-api-server-dep/configuring.mdx
+++ b/content/docs/run-api-server-dep/configuring.mdx
@@ -1,0 +1,39 @@
+---
+title: Configuring
+order: 30
+---
+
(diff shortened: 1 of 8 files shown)


rice2000 (approved):
LGTM!  Thank you.

PR #359, opened 2021-03-12
Add minor clarifications & cleanup to the Captive Core deployment docs

### LINKED ISSUE #330: Add missing `/metrics` info to Monitoring section
Add missing info from here: https://github.com/stellar/go/blob/master/services/horizon/internal/docs/reference/endpoints/metrics.md

To here:
https://developers.stellar.org/docs/run-api-server/monitoring/

Shaptic:
Highlights:
 - simplified commands
 - added system requirements
 - clarified how `/metrics` works (which closes #330).

### DIFF
--- a/content/docs/run-api-server/installing.mdx
+++ b/content/docs/run-api-server/installing.mdx
@@ -23,7 +23,7 @@ sudo apt update && sudo apt install stellar-core stellar-horizon
 
 </CodeExample>
 
-If you want to do [ingestion](./running.mdx#ingesting-transactions) on this machine, you should also install the `stellar-captive-core` package.  [This page](https://github.com/stellar/packages/blob/master/docs/installing-individual-packages.md#installing-individual-packages) outlines the various commands that these packages make available.
+[This page](https://github.com/stellar/packages/blob/master/docs/installing-individual-packages.md#installing-individual-packages) outlines the various commands that these packages make available.
 
 ### Building
 Should you decide not to use one of our prebuilt releases, you may instead build Horizon from source. To do so, you need to prepare a developer environment, including:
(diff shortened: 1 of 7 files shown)


content/docs/run-api-server/installing.mdx
```diff
@@ -21,69 +18,59 @@ SDF publishes new releases to its custom Ubuntu repositories. Follow [this guide
...
 
 ```bash
-sudo apt update && sudo apt install stellar-captive-core stellar-horizon
+sudo apt update && sudo apt install stellar-core stellar-horizon
 ```
 
 </CodeExample>
 
-[This page](https://github.com/stellar/packages/blob/master/docs/installing-individual-packages.md#installing-individual-packages) outlines the various commands that these packages make available.
+If you want to do [ingestion](./running.mdx#ingesting-transactions) on this machine, you should also install the `stellar-captive-core` package.  [This page](https://github.com/stellar/packages/blob/master/docs/installing-individual-packages.md#installing-individual-packages) outlines the various commands that these packages make available.
```
jacekn, line 26: `stellar-captive-core` is a configuration package. It configures core qorumset for use with captive core for convenience but it's not required. Some operators might choose to configure core themselves using config management, if they choose to do so it might even be better to avoid this package. @brahman81 should be able to confirm this.

jacekn: Also the link this paragraph points at does not contain information about stellar-captive-core

Shaptic: Ah, very interesting, thanks for the clarification! The link was intended to point to `stellar-core-cmd` and `stellar-horizon-cmd`, but I can see how that would be confusing. I've moved some things around in 0d944ff.

jacekn: Thanks, it looks great from my perspective now!

jacekn (approved):

rice2000 (approved):
These look great!  Thank you.

PR #362, opened 2021-03-17
Changed "it's" (it is) to "its" in a few places

slightknack:
I was scrolling through the docs and I noticed a few things referenced with "it's" that should've been "its". Nothing substantial, I know, but I'd like to help improve Stellar and this seems like a simple enough first step. :P


Shaptic (approved):
Awesome, ty!

PR #364, opened 2021-03-23
Fix a link and remove a redundant character in testnet.mdx

howardtw:
The link to re-join and re-sync the testnet was broken.


rice2000 (approved):
Nice catch!  Thank you!!!

PR #367, opened 2021-03-26
First pass on dropping Watcher references

Shaptic:
This drops the most obvious references to the now-deprecated Watcher nodes. 

There's still quite a lot to be done to remove all of the additional references that recommend or mention various ways to combine standalone Stellar Core nodes with Horizon (which we no longer want people doing due to Captive Core in Horizon 2.x, of course).


rice2000 (approved):
Nice.  I think this takes a lot of steps to update references, and will do a lot to avoid confusion and point people to the new architecture.  I'm all for merging this now and then looking for and addressing additional references to the deprecated architecture in another pass and PR.

PR #368, opened 2021-03-26
Clarify that fee bumps behave like normal TXs

Shaptic:
This should resolve any ambiguity about how fee-bump transactions should be submitted. (In other words, they aren't special and don't need an entry in the [API Reference](https://developers.stellar.org/api/).)


rice2000 (approved):
Nice.  Thanks!

PR #372, opened 2021-04-01
ensure installed packages are capped at anticipated versions

tyvdh:
Seeing a couple of these issues popping up: https://stellar.stackexchange.com/questions/4549/sass-error-undefined-variable and I'm fairly certain it's due to a version issue so I'm capping the version number on the problematic packages.


rice2000 (approved):
Thanks for fixing!

PR #375, opened 2021-04-05
Fixing typos, issuing vs supporting assets

lydiat:
A few typos, and adding a caveat about the language used for assets (we share this doc with anchors who also support assets)


rice2000 (approved):
Thank you!

Issue #376, opened 2021-04-07
Inconsistencies on Run a Core Node -> Overview - Watcher missing?


wanecek:
On the [Overview](https://developers.stellar.org/docs/run-core-node/)-page of Run a Core Node, the docs describe the different types of nodes that one can configure using stellar core.

There are however some inconsistencies between the total number of node types that the text refers to, and I am wondering if maybe the Watcher type is missing? 

### [Types of nodes](https://developers.stellar.org/docs/run-core-node/#types-of-nodes)

Under the section "Types of nodes", the following can be read (emphasis mine):

> To make things easier, we’ll define **four** types of nodes based on permutations of those two options: Basic Validator, Full Validator, and Archiver. You’ll notice that they all support Horizon and submit transactions to the network:

This is followed by a table listing **three** node types. Similarly, further down on the same page, under the subsection "Archiver", the final sentence reads:

> Otherwise, you’re probably better off choosing one of the other **three** types.

While only two other node types are discussed.

---

I'm quite new to the Stellar network, and am unsure of if the three nodes listed are what's what, or if the Watcher node type (e.g. discussed in the Lumenauts videos and shown on on Stellarbeat.io) has been deprecated as a concept?

wanecek:
Hmm, after finding reading the recent commit `[e8945ce2](https://github.com/stellar/new-docs/commit/e8945ce2cc245add925f6f74e0f816a05e75f488)` by @Shaptic, I believe it's only a matter of inconsistencies after editing the page to remove Watchers. I will submit a PR instead.

Shaptic:
Thank you! Great catches and contributions, appreciated :pray: I think we should acknowledge Watchers' existence and explain that they're deprecated, as opposed to hide them as I did in #367.

Issue #363, opened 2021-03-23
Documentation Wishlist Mega-Issue


Shaptic:
This is a potpourri of enhancements, improvements, fixes, etc. both large and small that are better off existing in a shared, collaborative space rather than the ephemeral notes and snippets scattered around desks everywhere. Anyone should feel free to take any of these up if they feel like contributing to the docs!

### Smaller Items
 - [ ] A description of `POST /transactions` (i.e. when you click it in the right-hand list) is missing in the [API Reference](https://developers.stellar.org/api/resources/transactions/post/)
 - [ ] Check if other resources have broken links (this should be scriptable! or at least script-assisted)
 - [ ] Remove Watcher references to the [Run a Core Node](https://developers.stellar.org/docs/run-core-node/) tutorial now that they're obsolete after Horizon 2.0
 - [ ] Ensure that the API reference is up-to-date and accurately reflects the current API

### Larger Items
 - [ ] A **glossary page** on path payments, to bridge the gap between Kolten's high-level overview](https://medium.com/stellar-community/understanding-stellar-path-payments-5eefe55b071b) and the [low-level details](https://developers.stellar.org/docs/start/list-of-operations/#path-payment-strict-send) on the Operations page.
 - [ ] A **glossary page** on pooled accounts, describing why & how they're used

tomerweller:
I'd like to emphasize the need for a tutorial regarding custodial/pooled accounts. We're getting a lot of inquiries on these.

PR #377, opened 2021-04-07
Fix inconsistent text after #367

wanecek:
This addresses a minor textual error on the Run a Core Node -> Overview page.

Specifically, after having removed the Watcher node type from the page, two
places in the text listed the wrong number of nodes (four instead of now three).

This addresses issue #376.

CC: @Shaptic


wanecek:
I'm wondering if it would be helpful to add a disclaimer of some sorts here saying that watcher nodes are now deprecated?

rice2000 (approved):
Good catch!  Thanks.  Really appreciate your contribution!

Shaptic:
@wanecek agreed! I've updated #363 accordingly.

PR #380, opened 2021-04-12
Add links to @marta-lokhova's knowledge shares about SCP voting

leighmcculloch:
### What
Add links to @marta-lokhova's knowledge shares about SCP voting.

### Why
Someone asked about how the voting works [here](https://lobste.rs/s/6lrojl/who_s_hiring_q2y2021#c_whh6vw) and I realized we don't have links to these great resources on the technical documentation page. They're worth including because I think they are more accessible for understanding how SCP works.


rice2000 (approved):
Nice addition!  Thank you.

Issue #304, opened 2020-12-18
Entered URL doesn't go to the right section


quietbits:
### What did you do?

Entered this URL in the browser: https://developers.stellar.org/api/errors/result-codes/operation-specific/change-trust/

### What did you expect to see?
I expected it to go to the "Change Trust Result Codes" section (Errors / Result Codes / Operation-Specific Result Codes)

### What did you see instead?
It landed on the wrong section (quite a bit lower than it should have).

vcarl:
(Removing "Docs" label because this is a technical detail, not a content issue)

I've dug into this a bit, and I"m not seeing an obvious cause for the jumps. It's possible this is an artifact of forced reflows caused by JS querying for the position of DOM nodes, it looks like this is even worse on Firefox.

One thing we might want to try (though I'm not confident it will help the issue) is the `history.scrollRestoration` api https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration

<img width="935" alt="Screen Shot 2020-12-21 at 4 41 02 PM" src="https://user-images.githubusercontent.com/1551487/102804514-c4811b80-43b1-11eb-9169-f2aa9d01941c.png">

See annotated flame chart. The first part of the deferred tasks after client JS loads seems to happen as the scroll position jumps, which _could_ be because one of the [`Route.js` effect dependencies](https://github.com/stellar/new-docs/blob/f506bdfb2499f1cf90f2480fa64a1671ff175d17/src/components/ApiReference/Route.js#L15-L20) is changing when it shouldn't

bartekn:
I just experienced it. I don't know how it works so my suggestion can be stupid but maybe a quickfix could be to try to go to a requested section again after 2-3 seconds after page load?

PR #381, opened 2021-04-13
Redirect to add trailing slash

paulbellamy:
Add trailing slashes on docs urls, to make sure you end up at the right place. I typed the url without a trailing slash and was suprised when it took me to the "wrong place".

For example:.
  https://developers.stellar.org/api/resources/effects/types/
    - takes you to the Effect Types section.
vs
  https://developers.stellar.org/api/resources/effects/types
    - Old behaviour: takes you to the top of the page
    - New behaviour: redirects to the above.

I opted to add slashes as that seems to be the convention we use in links.

I'm not an nginx wizard, so maybe there's a cleaner way to do the same behaviour?

(The other commit gets `yarn prod:build` working on osx, as the `date` util there doesn't have the `--rfc-3339` flag.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
 # If TAG is not provided set default value
 TAG ?= stellar/developers:$(shell git rev-parse --short HEAD)$(and $(shell git status -s),-dirty-$(shell id -u -n))
 # https://github.com/opencontainers/image-spec/blob/master/annotations.md
-BUILD_DATE := $(shell date --utc --rfc-3339=seconds)
+BUILD_DATE := $(shell date -u +%FT%TZ)
 
 docker-build:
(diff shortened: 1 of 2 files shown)


rice2000 (approved):
This makes a ton of sense to me, and seems to work. @vcarl or @quietbits  would you mind taking a quick look to verify this approach sits well with you?  My code vetting abilities are limited.

quietbits:
We have a related [open issue](https://github.com/stellar/new-docs/issues/304), and ending `/` doesn't seem to fix it, so most likely we'll need to fix it on the UI. But I'm OK with this fix if it helps with some cases, I just can't validate these changes (but they look OK to me). 😊 @vcarl I'll leave it up to your expertise.

vcarl:
This seems solid, can confirm that a trailing slash seems to improve behavior here

paulbellamy:
Turns out this 404'ed loads of stuff for some reason. reverted. Thanks @Shaptic

PR #387, opened 2021-04-20
Fix small typo in Fee Bump

raymens:


rice2000 (approved):
Good catch!  Thank you.

PR #388, opened 2021-04-21
Switch build image to ubuntu 20.04

jacekn:
* switch to ubuntu 20.04 base image
* move away from deprecated "apt-key add"
* add --pull to the Makefile to ensure we always use latest images
* add more google-chrome dependencies for 20.04


brahman81 (approved):
LGTM

quietbits (approved):

PR #395, opened 2021-04-28
Update decentralized-exchange.mdx

### LINKED ISSUE #394: Potential Editor Note Left in Docs
At the end of the first paragraph under [Passive Order](https://developers.stellar.org/docs/glossary/decentralized-exchange/), the voice changes parenthetically in a way that suggests this is an editors comment/question:

> Passive orders allow markets to have zero spread. If you want to exchange USD from anchor A for USD from anchor B at a 1:1 price, you can create two passive orders so the two orders don’t fill each other. (will this actually be a zero spread? it should still be a tick wide right? or would it be a “locked market”, where the bid and the ask is at the same price?)

rice2000:
Removing editorial aside.
closes #394

### DIFF
--- a/content/docs/glossary/decentralized-exchange.mdx
+++ b/content/docs/glossary/decentralized-exchange.mdx
@@ -34,7 +34,7 @@ For more information, take a look at [our guide on fees in Stellar](./fees.mdx).
 
 ## Passive Order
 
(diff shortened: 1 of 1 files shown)


Shaptic (approved):
:+1:

PR #402, opened 2021-05-03
Fix JSON responses in the API references to be valid JSON

Shaptic:
I'm currently in the process of writing an automated tool to assist with finding out-dated or invalid API reference documentation, and this is some of the preliminary results: invalid JSON in response examples.


rice2000 (approved):
LGTM!

Issue #374, opened 2021-04-04
Missing Diagram


malengatiger:
### What did you expect to see?

An orderbook can be summarized by a diagram, as shown below. It is often visible on the trading interface of an exchange, though sometimes inverted left-to-right or drawn horizontally. But the idea is the same.

rice2000:
closed by #405

rice2000:
Thanks for the tip, @malengatiger!

PR #405, opened 2021-05-05
Issue #374: missing diagram

ElliotFriend:
I've brought the diagram that was missing into the new-docs repo and inserted it into the same spot on the page as the old repo.


rice2000 (approved):
Well done!  Thanks for taking care of all of this.

PR #406, opened 2021-05-07
Adding DocSearch to /docs

quietbits:
Note: search links are absolute (pointing to production)

image

image


src/components/Search.js
```diff
@@ -0,0 +1,84 @@
...
+    outline-color: transparent;
+    outline-style: none;
+    border-color: ${PALETTE.purpleBlue};
+  }
+`;
+
+export const Search = () => {
+  useEffect(() => {
+    window.docsearch({
+      apiKey: "a90bee98943eff1ca1adfdaac1e434d8",
```
quietbits, line 64: Not hiding the key because it's visible in every request.

accordeiro: Yeah it's a public key, so all good. But maybe it's worth moving that to an external config variable, so it's easier to find / modify later?

accordeiro (approved):
Looks great! Excited about this feature! :) Left a small comment inline 👇

piyalbasu (approved):

PR #408, opened 2021-05-09
Update Go code for the claimable balance example

brettcannon:
Was missing a `check()` function, had an unnecessary import, and didn't handle two different return values.


Shaptic (approved):
Leaving off the `check` was actually intentional, so that example consumers thoughtfully implemented their own error-handling. However, there's clearly value in including it for completeness sake. Thanks for the change!

PR #413, opened 2021-05-13
Complete "Run a Core Node" transition to the new Captive Core model.

### LINKED ISSUE #376: Inconsistencies on Run a Core Node -> Overview - Watcher missing?

Shaptic:
Improves language surrounding Watcher nodes (elaborating on their deprecation) and ensures there are no extraneous references to the old architecture in this section.

Closes #376.

### DIFF
--- a/content/docs/run-core-node/configuring.mdx
+++ b/content/docs/run-core-node/configuring.mdx
@@ -5,7 +5,7 @@ order: 40
 
 import { CodeExample } from "components/CodeExample";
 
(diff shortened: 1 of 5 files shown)


rice2000 (approved):
Well done!  Thanks.

PR #415, opened 2021-05-13
Clarify how Horizon handles transaction resubmission.

Shaptic:
`POST /transactions` behaves differently on tx resub, and the API reference should reflect that.


rice2000 (approved):
Interesting.  I wasn't even aware of that behavior.  Makes sense!

PR #420, opened 2021-05-17
Update auth_revoke semantics

ire-and-curses:
[CAP 29](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0029.md) (deployed in protocol V16) was implemented: auth_required is no longer required to revoke.


rice2000 (approved):
Looks good!  Thanks! @Shaptic

PR #421, opened 2021-05-18
Removing expired images, tweaking some copy

tyvdh:


Shaptic (approved):
:+1: Any plans to replace the images, @tyvdh?

tyvdh:
Not in the short term. They weren't essential and I can't find them anywhere on my machine 😞

PR #422, opened 2021-05-19
Update Horizon's system requirements to mention a new parameter.

Shaptic:
It's possible to control the storage location of Captive Core's buckets as of v2.1, so the docs should reflect this.


rice2000 (approved):
Perfect.  Thanks @Shaptic!

PR #423, opened 2021-05-20
Update stellar-core installation instructions

jacekn:
* update docker image information
* packages now require Ubuntu 18.04 or later

### DIFF
--- a/content/docs/run-core-node/installation.mdx
+++ b/content/docs/run-core-node/installation.mdx
@@ -7,15 +7,34 @@ There are three ways to install Stellar Core: you can use a Docker image, use pr
 
 ## Docker-based installation
 
-SDF maintains a [quickstart image](https://github.com/stellar/docker-stellar-core-horizon) that bundles Stellar Core with Horizon and postgreSQL databases. It's a quick way to set up a default, non-validating, ephemeral configuration that should work for most developers.
+### Development environments
 
-The SDF also maintains a Stellar-Core-only [standalone image](https://github.com/stellar/docker-stellar-core) that starts a 3 node local stellar-core network, all on the same docker host
+SDF maintains a [quickstart image](https://github.com/stellar/docker-stellar-core-horizon) that bundles Stellar Core with Horizon and postgreSQL databases. It's a quick way to set up a default, non-validating, ephemeral configuration that should work for most developers.
 
 In addition to SDF images, Satoshipay maintains separate Docker images for [Stellar Core](https://github.com/satoshipay/docker-stellar-core) and [Horizon](https://github.com/satoshipay/docker-stellar-horizon). The Satoshipay Stellar Core Docker image comes in a few flavors, including one with the AWS CLI installed and one with the Google Cloud SDK installed. The Horizon image supports all Horizon environment variables.
 
+### Production environments
+
+The SDF also maintains a Stellar-Core-only [standalone image](https://hub.docker.com/repository/docker/stellar/stellar-core). The image utilizes deb packages so the stellar-core binary checksum can be verified.
+
+Example usage:
+
+```
+docker run stellar/stellar-core:latest help
+docker run stellar/stellar-core:latest gen-seed
+```
+
+To run daemon you need to provide a configuration file:
+```
+# Initialize postgres DB (see DATABASE config option)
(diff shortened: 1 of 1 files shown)


rice2000 (approved):
Great!  Thanks, @jacekn. @Shaptic PTAL when you get a chance.

content/docs/run-core-node/installation.mdx
```diff
@@ -7,15 +7,34 @@ There are three ways to install Stellar Core: you can use a Docker image, use pr
 
 ## Docker-based installation
 
-SDF maintains a [quickstart image](https://github.com/stellar/docker-stellar-core-horizon) that bundles Stellar Core with Horizon and postgreSQL databases. It's a quick way to set up a default, non-validating, ephemeral configuration that should work for most developers.
+### Development environments
 
-The SDF also maintains a Stellar-Core-only [standalone image](https://github.com/stellar/docker-stellar-core) that starts a 3 node local stellar-core network, all on the same docker host
+SDF maintains a [quickstart image](https://github.com/stellar/docker-stellar-core-horizon) that bundles Stellar Core with Horizon and postgreSQL databases. It's a quick way to set up a default, non-validating, ephemeral configuration that should work for most developers.
```
Shaptic, line 12: This is still the old Watcher model, right? i.e. Horizon + standalone Stellar-Core? I think we'd ideally we'd either (a) outline that this isn't recommended way to run Horizon or (b) talk about how to run the Core-only portion here.

content/docs/run-core-node/installation.mdx
```diff
@@ -7,15 +7,34 @@ There are three ways to install Stellar Core: you can use a Docker image, use pr
...
 
 In addition to SDF images, Satoshipay maintains separate Docker images for [Stellar Core](https://github.com/satoshipay/docker-stellar-core) and [Horizon](https://github.com/satoshipay/docker-stellar-horizon). The Satoshipay Stellar Core Docker image comes in a few flavors, including one with the AWS CLI installed and one with the Google Cloud SDK installed. The Horizon image supports all Horizon environment variables.
 
+### Production environments
+
+The SDF also maintains a Stellar-Core-only [standalone image](https://hub.docker.com/repository/docker/stellar/stellar-core). The image utilizes deb packages so the stellar-core binary checksum can be verified.
```
Shaptic, line 18: Should we add cross-references to the expected checksums?

content/docs/run-core-node/installation.mdx
```diff
@@ -7,15 +7,34 @@ There are three ways to install Stellar Core: you can use a Docker image, use pr
...
+
+Example usage:
+
+```
+docker run stellar/stellar-core:latest help
+docker run stellar/stellar-core:latest gen-seed
+```
+
+To run daemon you need to provide a configuration file:
```
Shaptic, line 27: Link to the [example configuration file](https://github.com/stellar/stellar-core/blob/master/docs/stellar-core_example.cfg) here?

Shaptic (commented):
Minor optional notes to address (will let you decide before merging), but otherwise LGTM! :+1:

content/docs/run-core-node/installation.mdx
jacekn: Good point. We don't publish checksums but we have cryptographically signed deb packages that we can use to verify the binary. I can probably come up with a one-liner to do that.

content/docs/run-core-node/installation.mdx
jacekn: Good idea. I'll need to think about it a bit more because simple in-container sqlite DB from the example you linked to won't work out of the box. But it's certainly a good idea to have something that works.

content/docs/run-core-node/installation.mdx
jacekn: Yes it is still the same image, we're not touching it at this point. I think the diff here is confusing, I did not touch this part of the document.

jacekn (commented):
> Minor optional notes to address (will let you decide before merging), but otherwise LGTM! +1

Thank you, you made some good points. I'll merge as is for now, I'd like to get the image published as soon as possible. But I'll definitely follow up with fixes.

PR #429, opened 2021-06-02
Uppercase (typo fix)

phocks:
Simple fix.


Shaptic (approved):
Thanks for the catch!

PR #433, opened 2021-06-10
Nitpick: Consistency in Tier 1 vs Tier One

wanecek:
Avoid mixing "Tier One" and "Tier 1".

_Hating myself just a little bit for nitpicking this, but.. yeah. Writing an article myself and was unsure which way Stellar prefers it to be written, maybe this avoids confusion._


rice2000 (approved):
Good point!  Thank you, @wanecek

PR #439, opened 2021-06-22
Fixing bug in the tutorial sample code

allanjsx:
In `txnbuild`, SourceAccount should take `txnbuild.Account` type instead of a `string`.


Shaptic (approved):
Good call, thanks!

PR #442, opened 2021-06-24
Side nav: update link colors

quietbits:


src/components/Documentation/Articles.js
```diff
@@ -91,6 +93,8 @@ const CustomList = styled(ListItem)`
     padding-left: ${(props) => (props.depth > 1 ? `${props.depth - 1}rem` : 0)};
     font-weight: ${(props) =>
       props.isActive ? FONT_WEIGHT.bold : FONT_WEIGHT.normal};
+
+    ${(props) => (props.isActive ? `color: ${PALETTE.purpleBlue}` : ``)};
```
piyalbasu, line 97: maybe we should break this out into a function that can be reused since we're doing it twice:

for ex:
```
const highlightActive = (isActive) => `color: ${PALETTE.purpleBlue}` : ``
```

quietbits: Good idea! 👍

piyalbasu (approved):

PR #462, opened 2021-07-14
Remove performance tracking

quietbits:
Removing performance tracking, which we no longer need. It was breaking the website in Firefox when `resistFingerprinting` flag was enabled.
Keeping the performance methods in case we'll need them again in the future.


Shaptic:
I can confirm that the preview docs no longer break in FF+Arch with `resistFingerprinting: true` set.

piyalbasu (approved):
This PR looks good. I've created this followup ticket because we should have a plan for what to do with performance tracking: https://github.com/stellar/product-engineering/issues/222

PR #464, opened 2021-07-15
Adding explicit redirect from older known_issues page

shanzzam:
We had a redirect from a previous old docs page pointing to the old docs repo in error. 
As a result, some users were seeing the deprecated github files while we saw some attempts to use https://developers.stellar.org/docs/software/known-issues.html which results in a 404. 

This PR redirects will redirect https://developers.stellar.org/docs/software/known-issues.html to https://github.com/stellar/go/issues.


shanzzam:
This may not be the ideal redirect but we don't have one file for 'known issue' with explanations like we [previously had](https://github.com/stellar-deprecated/docs/blob/master/software/known-issues.md)

vcarl (approved):

rice2000 (approved):
This looks good to me!  Thanks, @shanzzam.

PR #471, opened 2021-07-26
Add captive core storage path config

satyamz:
This will add a new flag `export CAPTIVE_CORE_STORAGE_PATH="/var/lib/stellar"` to horizon reingestion config. In the absence of this flag stellar-horizon would flood the current working directory with local buckets containing XDR files.

### DIFF
--- a/content/docs/run-api-server/ingestion.mdx
+++ b/content/docs/run-api-server/ingestion.mdx
@@ -145,6 +145,7 @@ export HISTORY_ARCHIVE_URLS=https://s3-eu-west-1.amazonaws.com/history.stellar.o
 export NETWORK_PASSPHRASE="Public Global Stellar Network ; September 2015"
 export STELLAR_CORE_BINARY_PATH=$(which stellar-core)
 export ENABLE_CAPTIVE_CORE_INGESTION=true
+export CAPTIVE_CORE_STORAGE_PATH="/var/lib/stellar"
 # Number of ledgers per job sent to the workers.
 # The larger the job, the better performance from Captive Core's perspective,
 # but, you want to choose a job size which maximizes the time all workers are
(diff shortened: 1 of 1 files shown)


jacekn (commented):
This looks good to me but I'd like the horizon team to confirm as well.

bartekn:
> In the absence of this flag stellar-horizon would flood the current working directory with local buckets containing XDR files.

Is it really a problem? In the long term we're planning to add reasonable defaults to most of the config values (https://github.com/stellar/go/issues/3719) so that users can start a working Horizon instance with only 2-3 config values. Users will be able to set the `CAPTIVE_CORE_STORAGE_PATH` if the don't like that Horizon stores files in the working dir..

satyamz:
>Is it really a problem? In the long term we're planning to add reasonable defaults to most of the config values (stellar/go#3719) so that users can start a working Horizon instance with only 2-3 config values. Users will be able to set the CAPTIVE_CORE_STORAGE_PATH if the don't like that Horizon stores files in the working dir.. 

@bartekn  
Not a big problem. I faced this problem when I ran re-ingestion without specifying `CAPTIVE_CORE_STORAGE_PATH` and my instance ran out of disk space. Its common that root partition is usually smaller and user running re-ingestion can face this issue.
@jacekn and I discussed this internally and we think we can add this option as a comment. WDYT?

ire-and-curses:
comment sounds great to me :+1:

bartekn:
> @jacekn and I discussed this internally and we think we can add this option as a comment. WDYT?

OK, sounds good.

satyamz:
Thank you @bartekn  and @ire-and-curses. I updated the PR. PTAL.

jacekn (commented):
Looks good from my perspective

ire-and-curses (approved):
LGTM

jacekn (approved):

bartekn (approved):

satyamz:
I am not able to merge this as I do not an access to merge. Can someone please merge this PR?

PR #481, opened 2021-07-28
Add Codeblock for listing all assets

MatejMecka:
This adds a Python snippet to list all asset with a name

### DIFF
--- a/content/api/resources/assets/list.mdx
+++ b/content/api/resources/assets/list.mdx
@@ -68,6 +68,16 @@ server
   });
 ```
 
+```python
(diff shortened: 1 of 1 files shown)


content/api/resources/assets/list.mdx
```diff
@@ -204,6 +214,17 @@ var es = server
   .cursor("now")
   .stream({ onmessage: callback });
 ```
+	
+```python
+
+def callback(resp):
+  print(resp)
```
Shaptic, line 223: spacing error?
```suggestion
    print(resp)
```

PR #483, opened 2021-07-28
Add Code Block for Querying a Claimable Balance in Python with an ID

MatejMecka:


content/api/resources/claimablebalances/single.mdx
```diff
@@ -58,6 +58,15 @@ server
   });
 ```
 
+```python
+
+try:
+	print(server.claimable_balances().claimable_balance("000000000102030000000000000000000000000000000000000000000000000000000000").call())
```
Shaptic, line 66: too long pls, try [this](https://stackoverflow.com/a/48528147):

```suggestion
print((server
     .claimable_balances()
     .claimable_balance("000000000102030000000000000000000000000000000000000000000000000000000000")
     .call()))
```

Shaptic (approved):
This is a bit of an edge case since it's almost impossible to make look better but the principle of trying to keep lines short still applies

PR #509, opened 2021-07-29
Page Arguments - Python Code Sample

MatejMecka:


content/api/introduction/pagination/page-arguments.mdx
```diff
@@ -57,6 +57,17 @@ server
     console.error(err);
   });
 ```
+  
+```python
+
+try:
+    print(server.operations().for_ledger("26478723").cursor("113725249324879872").limit(5).order("asc").call())
```
Shaptic, line 66: I would prefer not going over the line length, something like
```suggestion
    print(server.operations()
        .for_ledger("26478723")
        .cursor("113725249324879872")
        .limit(5)
        .order("asc")
        .call())
```
would look cleaner on the page.

PR #510, opened 2021-07-30
Adds a glossary entry describing muxed accounts

Shaptic:
This glossary entry is intended to be focused on the technical details behind muxed accounts, but may also be expanded to incorporate an ecosystem- and partner-level FAQ that provides adoption guidance rather than implementation and SDK usage examples.

Jump directly to a preview of the new page with [this link here](https://developers-pr510.previews.kube001.services.stellar-ops.com/docs/glossary/muxed-accounts/).

(cc: @tomerweller and @meiyib as additional reviewers)

### DIFF
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -5,15 +5,15 @@ order:
 
 import { Alert } from "components/Alert";
 
+
+A **muxed** (or *multiplexed*) **account** is an account that exists "on top of" a traditional Stellar account address. It combines the familiar `GABC...` address with a 64-bit integer ID and can be used to distinguish multiple "virtual" accounts that share an underlying "real" account.
+
 <Alert>
 
   **Warning**: *This feature is experimental.*
-  Multiplexed accounts are not widely supported yet. Many wallets may not display a muxed account address and instead display the account ID address, ignoring the 64-bit ID. Adoption efforts are actively in progress.
+  Muxed accounts are not widely supported yet. Many wallets may not display a muxed account address and instead display the account ID address, ignoring the 64-bit ID. Adoption efforts are actively in progress.
 
 </Alert>
-
-
-A **muxed** (or *multiplexed*) **account** is an account that exists "on top of" a traditional Stellar account address. It combines the familiar `GABC...` address with a 64-bit integer ID and can be used to distinguish multiple "virtual" accounts that share an underlying "real" account.
 
 Muxed accounts were defined in [CAP-27](https://stellar.org/protocol/cap-27), introduced in Protocol 13, and their string representation is described in [SEP-23](https://stellar.org/protocol/sep-21). They have their own address format that starts with an `M` prefix. For example, from the account address `GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ`, we can create new muxed accounts with two different IDs:
 
@@ -28,7 +28,7 @@ There are many other benefits to embedding this abstraction into the protocol:
 
   - **"Share-ability"**: Rather than worrying about error-prone things like copy-pasting memo IDs, you can just share your `M...` address.
   - **SDK support**: The various [SDKs](../software-and-sdks) support this abstraction natively, letting you create, manage, and work with muxed accounts easily.
-  - **Extensibility**: 64 bits available in the ID for distinction between accounts should allow any abstraction.
+  - **Efficiency**: By combining related virtual accounts under a single account's umbrella, you can avoid holding reserves and paying fees for all of them to exist in the ledger individually.
 
 <Alert>
 
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -81,11 +81,25 @@ Even though at the end of the day, only the underlying `G...` account _truly_ ex
 
 ## Examples
 
-In this section, we'll demonstrate how to create muxed accounts and how they interact with some of the supported operations. To drive home the fact that custodial account workarounds based on transaction memos are superfluous now, we'll use that as a skeleton for our example structure.
+In this section, we'll demonstrate how to create muxed accounts and how seamlessly they interface with their supported operations. To drive home the fact that custodial account workarounds based on transaction memos are superfluous now, we'll use that as a skeleton for our example structure.
+
+After preparing some supporting code, we'll demonstrate three scenarios:
+
+  - [normal](#), "full" Stellar account payments (i.e. G to G),
+  - [mixed](#) payments (i.e. M to G), and
+  - [fully muxed](#) payments (i.e. M to M)
+
+but use a shared function for all of them that does the real work, highlighting the ease of implementing muxed account support.
+
+<Alert>
+
+**Warning**: In the following code samples, proper error checking is omitted for brevity. However, you should always validate your results, as there are many ways that requests can fail. You can refer to the guide on [Handling Errors Gracefully](../tutorials/handling-errors.mdx) for tips on error management strategies.
+
+</Alert>
 
 
 ### Preamble
-First, let's create two accounts and then a handful of virtual accounts representing "custodial customers" that the parent account manages.
+First, let's create two accounts and then a handful of virtual accounts representing "custodial customers" that the parent account manages:
 
 <CodeExample title="Preamble">
 
@@ -118,6 +132,8 @@ async function preamble() {
 ```
 
 </CodeExample>
+
+We assume that these accounts exist on the testnet; you can replace them with your own keys and use the [friendbot](../tutorials/create-account.mdx#create-account) if you'd like.
 
 When we run this function, we'll see the similarity in muxed account addresses among the customers, highlighting the fact that they share a public key:
 
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -83,9 +83,9 @@ In this section, we'll demonstrate how to create muxed accounts and how seamless
 
 After preparing some supporting code, we'll demonstrate three scenarios:
 
-  - [normal](#), "full" Stellar account payments (i.e. G to G),
-  - [mixed](#) payments (i.e. M to G), and
-  - [fully muxed](#) payments (i.e. M to M)
+  - [normal](#payments), "full" Stellar account payments (i.e. G to G),
+  - [mixed](#muxed-to-unmuxed) payments (i.e. M to G), and
+  - [fully muxed](#muxed-to-muxed) payments (i.e. M to M)
 
 but use a shared function for all of them that does the real work, highlighting the ease of implementing muxed account support.
 
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -15,7 +15,7 @@ import { Alert } from "components/Alert";
 
 A **multiplexed** (or **muxed**) **account** is an account that exists "on top of" a traditional Stellar account address. It combines the familiar `GABC...` address alongside a 64-bit integer ID in order to support differentiating multiple "virtual" accounts that share an underlying "real" account.
 
-They were introduced into the protocol in [CAP-21](https://stellar.org/protocol/cap-21) and their string representation is described in [SEP-23](https://stellar.org/protocol/sep-21). Muxed accounts have their own address format that starts with an `M` prefix. For example, from the account address `GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ`, we can create two new muxed accounts with different IDs:
+They were defined in [CAP-27](https://stellar.org/protocol/cap-27) and introduced into Stellar protocol in protocol version 13. Their string representation is described in [SEP-23](https://stellar.org/protocol/sep-21). Muxed accounts have their own address format that starts with an `M` prefix. For example, from the account address `GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ`, we can create two muxed accounts with different IDs:
 
   - `MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAAAAAAAACJUQ` has the ID `0`, while 
   - `MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAAAAAABUTGI4` has the ID `420`.
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -23,7 +23,7 @@ Muxed accounts were defined in [CAP-27](https://stellar.org/protocol/cap-27), in
 
 Both of these addresses, when used with one of the [supported operations][supported-ops], will act on the underlying `GA7Q...` address.
 
-With muxed accounts, we can *natively* support the ecosystem-level abstraction of account memo requirements (see [SEP-29](https://stellar.org/protocol/sep-29)), where transaction memos were often used to distinguish between users backed by a single custodial account. Now, the embedded ID can be used for this distinction, and guides like [this one](../building-apps/setup-custodial-account.mdx) can be made superfluous.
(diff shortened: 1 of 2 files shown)


content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+<Alert>
+
+  **Warning**: *This feature is experimental.*
+  Multiplexed accounts are not widely supported yet. Many wallets may not display a muxed account address and instead display the account ID address, ignoring the 64-bit ID. Adoption efforts are actively in progress.
+
+</Alert>
+
+
+A **multiplexed** (or **muxed**) **account** is an account that exists "on top of" a traditional Stellar account address. It combines the familiar `GABC...` address alongside a 64-bit integer ID in order to support differentiating multiple "virtual" accounts that share an underlying "real" account.
```
leighmcculloch, line 16:
```suggestion
A **multiplexed** (or **muxed**) **account** is an account that exists "on top of" a traditional Stellar account address. It combines the familiar `GABC...` address with a 64-bit integer ID and can be used to distinguish multiple "virtual" accounts that share an underlying "real" account.
```

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+They were introduced into the protocol in [CAP-21](https://stellar.org/protocol/cap-21) and their string representation is described in [SEP-23](https://stellar.org/protocol/sep-21). Muxed accounts have their own address format that starts with an `M` prefix. For example, from the account address `GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ`, we can create two new muxed accounts with different IDs:
```
leighmcculloch, line 18:
```suggestion
They were defined in [CAP-27](https://stellar.org/protocol/cap-27) and introduced into Stellar protocol in protocol version 13. Their string representation is described in [SEP-23](https://stellar.org/protocol/sep-21). Muxed accounts have their own address format that starts with an `M` prefix. For example, from the account address `GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ`, we can create two muxed accounts with different IDs:
```

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+  - `MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAAAAAAAACJUQ` has the ID `0`, while 
+  - `MA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJUAAAAAAAAAABUTGI4` has the ID `420`.
+
+Both of these addresses, when used with one of the [supported operations][supported-ops], will act on behalf of their underlying `GA7Q...` address.
```
leighmcculloch, line 23:
```suggestion
Both of these addresses, when used with one of the [supported operations][supported-ops], will act on the underlying `GA7Q...` address.
```

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+With muxed accounts, we can *natively* support the ecosystem-level abstraction of account memo requirements (see [SEP-29](https://stellar.org/protocol/sep-29)), where the memo was often used to differentiate between customers backed by a single custodial account. Now, the embedded ID can be used for this distinction, and guides like [this one](../building-apps/setup-custodial-account.mdx) can be made superfluous.
```
leighmcculloch, line 25:
```suggestion
With muxed accounts, we can *natively* support the ecosystem-level abstraction of account memo requirements (see [SEP-29](https://stellar.org/protocol/sep-29)), where the memo was often used to distinguish between customers backed by a single custodial account. Now, the embedded ID can be used for this distinction, and guides like [this one](../building-apps/setup-custodial-account.mdx) can be made superfluous.
```

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+There are many other benefits to embedding this abstraction into the protocol:
+
+  - **"Share-ability"**: Rather than worrying about error-prone things like transaction memo IDs, you can just share the `M...` address.
+  - **SDK support**: The various [SDKs](../software-and-sdks) support this abstraction natively, letting you create, manage, and work with muxed accounts easily.
+  - **extensibility**: Using 64 bits for the integer ID embedded into each muxed account ensures that most abstractions can be encoded; it can be used to represent almost anything.
```
leighmcculloch, line 31: This piece on extensibility is unclear to me, we made a mistake in making these IDs only 64bits and not arbitrary lengths, or at least 128bits because how M addresses are defined doesn't allow us to extend them. I might be misunderstanding the intent with discussing extensibility here. Could you share an example of what you mean?

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+<Alert>
+
+  **Terminology Tip**: The term *multiplexing* (often contracted to *muxing*) comes from computer networking and telecommunications, where multiple signals are combined into one signal over a shared medium.
+  We are doing the same thing with muxed accounts: combining multiple "virtual" accounts into a single shared account in the Stellar ledger.
+
+</Alert>
+
+
+It's worth noting that this feature is intended to be a high-level abstraction, merely embedded into the protocol for convenience and standardization. As far as the Stellar Network is concerned, all of the [supported operations][supported-ops] operate exactly as if you did not used a muxed account. For example, if you make a payment from two muxed accounts that share an underlying Stellar account (but have different IDs), this is *exactly the same* as that Stellar account sending two payments as far as the ledger is concerned.
```
leighmcculloch, line 42: We could also call out that there is no validation of muxed account ID's. Maybe unnecessary.
```suggestion
It's worth noting that this feature is intended to be a high-level abstraction, merely embedded into the protocol for convenience and standardization. As far as the Stellar Network is concerned, all of the [supported operations][supported-ops] operate exactly as if you did not use a muxed account. For example, if you make a payment from two muxed accounts that share an underlying Stellar account (but have different IDs), this is *exactly the same* as that Stellar account sending two payments as far as the ledger is concerned.
```

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+At the end of the day, only the underlying `G...` account _truly_ exists.
```
leighmcculloch, line 44:
```suggestion
At the end of the day, only the underlying `G...` account _truly_ exists on the ledger. 
```

content/docs/glossary/muxed-accounts.mdx
leighmcculloch, line 16: The term 'muxed' is well defined to be 'multiplexed' even in regular dictionaries. Given that we're calling them muxed accounts in SDKs and resolved to use that name for them, can we call them only muxed accounts rather than use both names in the doc?

leighmcculloch (commented):
This is great, suggestions throughout, but I defer to you and others on all. I really like the connection to SEP-29 too.

content/docs/glossary/muxed-accounts.mdx
Shaptic: ooh yes good catch haha

content/docs/glossary/muxed-accounts.mdx
Shaptic: Honestly kinda phoned it in for this one; I wanted a >= 3 item list. The idea was that 64 bits should be plenty for any sort of "virtual encoding" users might want. Can you think of any other perks to include of muxed accounts instead of this item?

content/docs/glossary/muxed-accounts.mdx
Shaptic: That seems fine with me, though from a cursory glance it seemed like the definition of "muxed" usually uses "multiplex" to define itself (super helpful :roll_eyes:). Notice that I included the "full" version of it just once here and use "muxed" exclusively since that's consistent with our SDKs, etc. as you mentioned and I agree that that's how we should refer to it. I just thought it'd be nice to at least associate the two words together for familiarity. Maybe switching the order ("**muxed** (or *multiplexed*) **account** is...") would be better?

Ultimately, though, I didn't participate in the naming discussions y'all had, so I'll follow whatever the consensus is!

leighmcculloch: Good point. I think lead with muxed, and move multiplexec into the brackets is the right call. Keeping the bolded text the thing we want people to call it +1.

content/docs/glossary/muxed-accounts.mdx
leighmcculloch: Unfortunately 64 bits was in some ways a mistake and we can't actually encode any virtual encoding into it. 128bits would have been better, but here we are.

Opaque is a big benefit that helps with shareability and you could mention that.

Shaptic: I thought of another perk; the bullet is now "efficiency." Let me know what you think!

Shaptic: What did you mean by "opaque"?

content/docs/glossary/muxed-accounts.mdx
Shaptic: Done!

content/docs/glossary/muxed-accounts.mdx
leighmcculloch: Efficiency sgtm. I think you can mention that virtual accounts are now first class citizens.

Opaque in the sense the muxed account is just a single address even though it contains the elements described in this doc, from much of the ecosystems pov it's just another account and what's inside it is irrelevant.

content/docs/glossary/muxed-accounts.mdx
Shaptic: Added a blurb re: validation, good idea!

content/docs/glossary/muxed-accounts.mdx
Shaptic: Good call re: first class! I think the opaque notion is good but it's in some ways covered by the other bullets, so efficiency fits better here.

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+</Alert>
+
+It's worth noting that this feature is intended to be a high-level abstraction, merely embedded into the protocol for convenience and standardization. There's no validation on IDs: as far as the Stellar Network is concerned, all of the [supported operations][supported-ops] operate exactly as if you did not used a muxed account. For example, if you make a payment from two muxed accounts that share an underlying Stellar account (but have different IDs), this is *exactly the same* as that Stellar account sending two payments as far as the ledger is concerned.
+
+It's worth noting that this feature is intended to be a high-level abstraction, embedded into the protocol for convenience and standardization. As far as the Stellar Network is concerned, all of the [supported operations][supported-ops] operate exactly as if you did *not* use a muxed account. For example, if you make a payment from two "sibling" muxed accounts (i.e. muxed accounts sharing an underlying `G...` account but with different IDs), this is *exactly the same* as that Stellar account sending two payments, as far as the ledger is concerned.
```
quietbits, line 42: These two paragraphs are almost exactly the same.

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+const custodian = sdk.Keypair.fromSecret("SAQLZCQA6AYUXK6JSKVPJ2MZ5K5IIABJOEQIG4RVBHX4PG2KMRKWXCHJ");
+const outsider  = sdk.Keypair.fromSecret("SAAY2H7SANIS3JLFBFPLJRTYNLUYH4UTROIKRVFI4FEYV4LDW5Y7HDZ4");
+
+async function preamble() {
+  custodianAcc = await server.loadAccount(custodian.publicKey());
+  outsiderAcc  = await server.loadAccount(outsider.publicKey());
+
+  customers = ["1", "22", "333", "4444"].map(
+    (id) => custodianAcc.createSubaccount(id)
```
quietbits, line 100: Is this the preferred way of creating a muxed account (and not `new MuxedAccount()`)?

quietbits (commented):
I like the background info about the muxed accounts. 🙌 Just left a couple of small comments.

content/docs/glossary/muxed-accounts.mdx
shanzzam: +1 on this question. I remember some discussion on this when implementing support in the Lab or Account Viewer. Are there differences/value  in the use of either method?

shanzzam (commented):
This is a great guide, nice work! Just a couple small comments.

content/docs/glossary/muxed-accounts.mdx
Shaptic: Oops, this must've been a merge mistake... thank you!

content/docs/glossary/muxed-accounts.mdx
Shaptic: Good question! Suppose `account` is some Stellar account loaded from Horizon. Then, these two are functionally equivalent:

```js
let muxed1 = new MuxedAcccount(account, "1");
let muxed2 = account.createSubaccount("1");
```

These are also functionally equivalent (in situations where you only have a muxed account variable around):

```js
let muxed3 = new MuxedAccount(muxed1.baseAccount(), "3");
let muxed4 = muxed1.createSubaccount("3");
```

The `createSubaccount()` method was created with this latter scenario in mind: conveniently create "sibling" muxed accounts from existing ones. It was added to regular `Account`s for interface compatibility (though the [naming has some contention](https://github.com/stellar/js-stellar-base/issues/433)). 

-------------------

Given that this question came up, maybe this is confusing? Perhaps having two ways to do the same thing is a bad move; the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) comes to mind here:  

> There should be one-- and preferably only one --obvious way to do it.

Thoughts on dropping this `createSubaccount()` helper, then?

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,305 @@
+---
+title: Muxed Accounts
+order:
+---
+
+import { Alert } from "components/Alert";
+import { CodeExample } from "components/CodeExample";
+
+
+A **muxed** (or *multiplexed*) **account** is an account that exists "on top of" a traditional Stellar account address. It combines the familiar `GABC...` address with a 64-bit integer ID and can be used to distinguish multiple "virtual" accounts that share an underlying "real" account.
```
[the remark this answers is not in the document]
Shaptic: Good point! Especially considering I think I use "parent" when referring to traditional accounts. I'll revise this.

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -0,0 +1,305 @@
...
+
+It's worth noting that this feature is intended to be a high-level abstraction, embedded into the protocol for convenience and standardization. As far as the Stellar Network is concerned, all of the [supported operations][supported-ops] operate exactly as if you did *not* use a muxed account. For example, if you make a payment from two "sibling" muxed accounts (i.e. muxed accounts sharing an underlying `G...` account but with different IDs), this is *exactly the same* as that Stellar account sending two payments, as far as the ledger is concerned.
+
+
+## Supporting Opt-In
+
+Because muxed account support is still experimental, you need to **explicitly opt-in** to creating operations with muxed properties. If you do not pass these opt-in flags in the necessary places, SDKs will throw errors. This is done in an effort to ensure that muxed accounts are not "accidentally" working in your application and thus may be misleading to your users.
+
+Each SDK implements this a little differently, but generally-speaking you need to explicitly call out muxing support when creating operations and/or transactions that use muxed accounts.
```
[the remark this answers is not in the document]
Shaptic: It's set for every op/tx, yeah. Should I add in "each" in this last sentence to drive that home?

> generally-speaking you need to explicitly call out muxing support when creating each operation and/or transaction that uses muxed accounts.

Or do you think it needs more clarity?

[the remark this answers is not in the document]
shanzzam: > Should I add in "each" in this last sentence to drive that home?
I think that's all that's needed. 

Side question: will the flag remain in perpetuity or is this a temporary setting?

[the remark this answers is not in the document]
Shaptic: I don't know the answer to that, honestly. If I had to surmise, it'll be a two-phase process: first, `withMuxing: true` will be the default (i.e. opt-**out** instead of opt-in). Then, we'll be able to drop the flag entirely.

content/docs/glossary/muxed-accounts.mdx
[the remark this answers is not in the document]
Shaptic: Done in 248b0d6; thoughts?

rice2000 (approved):
Nice!

content/docs/glossary/muxed-accounts.mdx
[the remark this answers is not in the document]
shanzzam: Clearer wording, thanks!

PR #585, opened 2021-11-02
doc - update send-and-receive-payments.mdx

Jonas-SRB:
The given Python example in the current version contains a wrong attribute for the .append_payment_op method within the TransactionBuilder. I fixed it by creating an Asset directly from the Asset Model and added the required import.


content/docs/tutorials/send-and-receive-payments.mdx
```diff
@@ -239,7 +240,7 @@ transaction = (
     )
         # Because Stellar allows transaction in many currencies, you must specify the asset type.
         # Here we are sending Lumens.
-        .append_payment_op(destination=destination_id, amount="10", asset_code="XLM")
+        .append_payment_op(destination=destination_id, amount="10", asset=Asset("XLM"))
```
Shaptic, line 243: Note we can also do `Asset.native()` here.

Shaptic (approved):
Nice catch, looks like the signature changed in the latest SDK release.

PR #591, opened 2021-11-12
Add context for CHANGE_TRUST_INVALID_LIMIT error

JakeUrban:
A partner mentioned how this wasn't clear in the docs. Although its not semantically different, removing a trustline is conceptually different than adjusting it's limit, so having a invalid limit error raised when trying to remove it entirely can be confusing.


rice2000 (approved):
Definitely clearer.  Thank you!

PR #598, opened 2021-11-18
Moving /api to new archive_content

shanzzam:
This should remove api from Gitbook content. 
This will give us the ability to replace with a link to current API docs.


Shaptic (approved):
We don't really need to worry about deleting stuff heh but :+1:

PR #610, opened 2021-12-15
Update how-to-issue-an-asset.mdx

BLevine11:
how to issue asset should not issue native asset


Shaptic (approved):
Good point, thank you!

PR #612, opened 2021-12-20
Updating Prereqs page w HW specs

jcx120:
adding emphasis on using Captive Core by default and reference HW specs for historical and daily ingestion based on benchmarks. Fixes #592

PR #616, opened 2021-12-21
Break down Horizon system requirements based on usage scenario

Shaptic:
This replaces #612 after some updates and polish were added to the original draft.


rice2000 (approved):

jcx120:
Looks good @Shaptic. Minor nit but would be nice to add links to scaling page for references to standalone and enterprise topologies in the pre-req page.

PR #618, opened 2022-01-01
Update build deps to use newer sharp and puppeteer libs 

sreuland:
Problem: Tried to run the web server locally for verifying development on arm64 macboook, some binary libs from deps didn't work, such as sharp and puppeteer.

Solution: Updated build deps to use newer sharp and puppeteer libs that support arm64 runtime.


Shaptic:
@quietbits can you either review this or @ someone who could? :smile:

quietbits (approved):
The update looks good and nothing broke. :raised_hands: :crossed_fingers:

sreuland:
@Shaptic , would you mind merging this? I don't have auth on repo, thanks!

PR #619, opened 2022-01-04
Add an FAQ section for muxed account rollout

Shaptic:
The covered questions are:

1. Sending to the wrong account:
  a. Your operations use muxed accounts, but the recipients do not support muxed accounts.
  b. You want to send to a muxed account, but your platform does not support them.
2. Your platform receives a transaction with muxed destination account(s) and a memo ID.
3. Using a muxed address without enabling the feature flag

Please suggest more.

### DIFF
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -177,8 +177,7 @@ function doPayment(source, dest) {
         source: source.accountId(),
         destination: dest.accountId(),
         asset: sdk.Asset.native(),
-        amount: "10",
-        withMuxing: true,
+        amount: "10"
       });
 
       let tx = new sdk.TransactionBuilder(accountBeforePayment, {
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -297,6 +297,16 @@ In this case, do not use a muxed address. The platform will likely fail to creat
 
 In an ideal world, this situation would never happen. You can determine whether or not the underlying IDs are equal; if they aren't, this is a malformed transaction and we recommend not submitting it to the network.
 
+### What happens if I get errors when using muxed accounts?
+
+In up-to-date versions of Stellar SDKs, muxed accounts are natively supported by default. If you are using an older version of an SDK, however, they may still be hidden behind a feature flag. 
+
+If you get errors when using muxed addresses on [supported operations][supported-ops] like:
+
+> destination is invalid; did you enable muxing?
+
+We recommend upgrading to the latest version of any and all Stellar SDKs you use. However, if that's not possible for some reason, you can will need to enable the feature flag before interacting with muxed accounts. Consult your SDK's documentation for details.
+
 ### What happens if I pass a muxed address to an incompatible operation?
 Only certain operations allow muxed accounts, as described [above][supported-ops]. Passing a muxed address to an incompatible parameter with an up-to-date SDK _should_ result in a compilation or runtime error at the time of use.
 
--- a/content/docs/glossary/muxed-accounts.mdx
+++ b/content/docs/glossary/muxed-accounts.mdx
@@ -1,3 +1,4 @@
+
 ---
 title: Muxed Accounts
 order:
@@ -11,8 +12,8 @@ A **muxed** (or *multiplexed*) **account** is an account that exists "virtually"
 
 <Alert>
 
-  **Warning**: *This feature is experimental.*
-  Muxed accounts are not widely supported yet. Many wallets may not display a muxed account address and instead display the account ID address, ignoring the 64-bit ID. Adoption efforts are actively in progress.
+  **Warning**: *This feature is in active rollout.*
+  Muxed accounts are gaining adoption but may not be fully supported. Some wallets may not display a muxed account address and instead display the account ID address, ignoring the 64-bit ID. Continued adoption efforts are actively in progress.
 
 </Alert>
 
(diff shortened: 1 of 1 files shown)


content/docs/glossary/muxed-accounts.mdx
```diff
@@ -59,6 +59,7 @@ Not all operations can be used with muxed accounts. For example, you cannot set
     * [`PathPaymentStrictSend`](../start/list-of-operations.mdx#path-payment-strict-send), and
     * [`PathPaymentStrictReceive`](../start/list-of-operations.mdx#path-payment-strict-receive)
   - the destination of an [`AccountMerge`](../start/list-of-operations.mdx#account-merge), and
+  - the "target" of a [`Clawback`](../start/list-of-operations.mdx#clawback) operation (i.e. the `from` field)
```
leighmcculloch, line 62: 💡 We don't use double quotes in this fashion elsewhere in the docs, so I'd avoid using it here unless you think it adds value.
```suggestion
  - the target of a [`Clawback`](../start/list-of-operations.mdx#clawback) operation (i.e. the `from` field) 
```

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -272,4 +273,81 @@ If we were to make a payment between two muxed accounts that had *different* und
 As is the case for most protocol-level features, you can find more usage examples and inspiration in the relevant test suite for your favorite SDK. For example, [here](https://github.com/stellar/js-stellar-base/blob/master/test/unit/muxed_account_test.js) are some of the JavaScript test cases.
 
 
+## Frequently-Asked Questions
```
leighmcculloch, line 276: 💭 Each of these sections don't pose a question. If we want this to be a FAQ we should probably pose them as Q and As. e.g. "What happens if a sender sends amounts to a muxed account but the receiver doesn't use muxed accounts?"

Maybe the title shouldn't be FAQ since this doesn't look like a FAQ.

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -272,4 +273,81 @@ If we were to make a payment between two muxed accounts that had *different* und
...
+
+</CodeExample>
+
+the runtime result would be:
+
+> Error: invalid version byte. expected 48, got 96
+
+This error message indicates that the `trustor` failed to parse as an account ID (`G...`). In other words, your code will fail and the invalid operation will never reach the network.
+
+### Using a muxed address without enabling the feature flag
```
leighmcculloch, line 328: 💭 We're about to enable the feature flag by default, so maybe we don't need this section? See https://stellarfoundation.slack.com/archives/C020KGHM5KQ/p1641402959045800

leighmcculloch (commented):
Some thoughts and suggestions, see inline.

content/docs/glossary/muxed-accounts.mdx
rice2000: Maybe the title should be something like "Common Issues Using Muxed Accounts" or something like that.

content/docs/glossary/muxed-accounts.mdx
rice2000: We may want to keep this in in case someone is using an old version of an SDK.  But I think we should acknowledge that by adding a bit of copy that says something like "In up-to-date versions of Stellar SDKs, muxed accounts are on by default, but if you are using an earlier version of an SDK, they may still be behind a feature flag.  Our first  recommendation if you get this error is to install the newest version of any and all Stellar SDKs you use.  However, if that's not possible for some reason, you can will need to enable the feature flag before interacting with muxed accounts."

content/docs/glossary/muxed-accounts.mdx
Shaptic: Consistency :+1: - alleviated.

content/docs/glossary/muxed-accounts.mdx
Shaptic: I've amended the headings to pose questions!

content/docs/glossary/muxed-accounts.mdx
Shaptic: Good idea, done!

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -221,7 +216,7 @@ preamble
 
 </CodeExample>
 
-Notice that we still sign the transaction with the `custodian` keys, because muxed accounts have no concept of secret keys. Ultimately, everything still goes through the "parent" account, and so we should see the account's balance decrease by 10 XLM accordingly:
+Notice that we still sign the transaction with the `custodian` keys, because **muxed accounts have no concept of a secret key**. Ultimately, everything still goes through the "parent" account, and so we should see the account's balance decrease by 10 XLM accordingly:
```
rice2000, line 219:
```suggestion
Notice that we still sign the transaction with the `custodian` keys, because **muxed accounts have no concept of a secret key**. Ultimately, everything still goes through the "parent" account, and so we should see the parent account's balance decrease by 10 XLM accordingly:
```

content/docs/glossary/muxed-accounts.mdx
```diff
@@ -272,4 +267,82 @@ If we were to make a payment between two muxed accounts that had *different* und
...
+
+### What happens if I get errors when using muxed accounts?
+
+In up-to-date versions of Stellar SDKs, muxed accounts are natively supported by default. If you are using an older version of an SDK, however, they may still be hidden behind a feature flag. 
+
+If you get errors when using muxed addresses on [supported operations][supported-ops] like:
+
+> destination is invalid; did you enable muxing?
+
+We recommend upgrading to the latest version of any and all Stellar SDKs you use. However, if that's not possible for some reason, you can will need to enable the feature flag before interacting with muxed accounts. Consult your SDK's documentation for details.
```
rice2000, line 308:
```suggestion
We recommend upgrading to the latest version of any and all Stellar SDKs you use. However, if that's not possible for some reason, you will need to enable the feature flag before interacting with muxed accounts. Consult your SDK's documentation for details.
```

rice2000 (approved):
2 small changes.  Other than that, I think it's ready to merge.

PR #632, opened 2022-01-26
Update testnet.mdx

rice2000:
- add 2022 testnet reset dates
- add a bit of info about Friendbot and rate limiting


Shaptic (approved):
Nice!

PR #636, opened 2022-01-28
Update Horizon deployment docs to reflect the latest release

Shaptic:
- ingestion is enabled by default
 - guidance around building stellar-core
 - there is no stellar-captive-core package on stable
 - more guidance around using pubnet

cc @stellar/horizon-committers @rice2000


jcx120 (commented):
For the Configuration doc, this might need some clarification: 

When running a non-ingesting Horizon, will Horizon still have a local Captive Core enabled by default that is watching the network that it can point to for the `--stellar-core-url ` config entry to or should this be some external Validator node pulled from a pubnet config file (T1 etc)?  References to Remote Captive Core, since its not well supported might confuse new entrants, ie in this sentence:

> This can be a Remote Captive Core instance with its underlying Core node exposed or a standalone Stellar-Core instance.

Overall, need clarity on setting up Horizon instances that are only for serving API requests that are non-ingesting and how they should be configured (if this is what this section is geared towards).

rice2000 (approved):
Nice.

kryptozubmixtape (approved):

PR #656, opened 2022-03-17
update withdrawLiquidity

alejomendoza:


pangsirirat422 (approved):

sasanpc (approved):
s

rezalashani1353 (approved):
#655

Shaptic:
It seems like this is just running the Markdown file through a particular flavor of JS and MD linting. Since different linter configs prefer different styles, I don't think this is a change worth merging.

PR #658, opened 2022-03-28
Adding Algolia Autocomplete search bar to /docs

alessandrak:
Adding Algolia Autocomplete search bar to /docs with customized content preview.

The `AutocompleSearch` component uses Algolia css and custom css, which is in the `static/autocomplete-stellar.css` file.
All keys are in `config.js` file and are public and production keys.

Screen size >= 680px
Top
Screen Shot 2022-03-28 at 09 50 09
Bottom
Screen Shot 2022-03-28 at 10 13 01

Screen size < 680px
Screen Shot 2022-03-28 at 10 10 48


static/autocomplete-stellar.css
```diff
@@ -0,0 +1,250 @@
...
+  max-width: 900px;
+  width: 100%;
+  overflow: none;
+  height: 485px;
+  max-height: 485px;
+}
+
+.aa-Panel button {
+  background-color: rgba(62, 27, 219, 0.04);
+  padding-top: 8px;
```
piyalbasu, line 11: Let's use `rem` for things like padding, margin, and font size

src/constants/config.js
```diff
@@ -1,2 +1,7 @@
 export const DOCSEARCH_DOCS_KEY = "a90bee98943eff1ca1adfdaac1e434d8";
 export const DOCSEARCH_DOCS_INDEX_NAME = "stellar";
```
quietbits, line 2: I think we can remove these now.

quietbits (commented):
New search looks good! :tada: Could you please remove everything from the old search (DocSearch)? That way it won't be confusing which one is used and avoid potential conflicts with the new one. :pray:

quietbits (approved):

piyalbasu (approved):

PR #686, opened 2022-06-14
multisig: use "Alternative Signer Types" instead of "Alternate Signature Types"

howardtw:
After talking to @rice2000 , we thought it is better to use the more common word `Alternative` to avoid confusion.


rice2000 (approved):
Thanks!

PR #691, opened 2022-07-11
Remove questionnaire link

briwylde08:
Tom CL wanted to remove the questionnaire link because he's getting submissions without the correct information/context.


rice2000 (approved):
This looks good, and I think we can go ahead and merge to defray any incoming.  If we decide to add it back with more context later, we can do that in a separate PR.  Nice job!

PR #696, opened 2022-07-27
In connect to anchors section:  - Removed any mention to stellar-anchor-server.herokuapp.com since it is not used anymore and available for takeover.

C0x41lch0x41:
HackerOne report: 1649820


C0x41lch0x41:
@rice2000 @JakeUrban please take a look thanks!

JakeUrban (approved):
We could use the testanchor.stellar.org domain, but if we choose not to, this LGTM.

C0x41lch0x41:
Ah did not know we had this domain. That may be better. Let me push an update.

rice2000 (approved):
Nice!

PR #704, opened 2022-08-24
clarify op types for payments

JakeUrban:
The docs list a set of operations that are returned from the `/payments` endpoint. However, it listed a `path_payment` operation, which was replaced with the strict send & strict receive ops awhile ago.


rice2000 (approved):
Thanks!

paulbellamy (approved):

PR #706, opened 2022-11-19
Fix API service example for minimal captive-core

leighmcculloch:
### What
Add the `NETWORK_PASSPHRASE` parameter to the minimal captive-core config example.

### Why
Running stellar-core in captive core mode without the network passphrase in the config results in an error:
```2022-11-18T16:42:58.430 GBATA [default FATAL] Got an exception: NETWORK_PASSPHRASE not configured```


Shaptic (approved):

leighmcculloch:
@Shaptic Could you merge if you have merge ability? I do not have it.

